PR #11, opened 2026-02-25
feat: add renderCall and renderResult for MCP tools

titouanmathis:
### `renderCall` — inline display of tool invocation
Both direct tools and the `mcp` proxy tool now display their name and arguments inline when the LLM calls them, matching the style of native pi tools like `read` and `edit`:

```
productive resource=tasks action=list filter={"assignee_id":"500521"}
mcp search: "query"
mcp tool_name {"key": "value"}
```

### `renderResult` — compact result with Ctrl+O expand
Results are shown collapsed to 10 lines by default, with a native `... (N more lines, ctrl+o to expand)` hint — consistent with how `bash`, `read`, and other built-in tools behave.

### Cleanup
- Removed all `console.log` calls for MCP tool invocations

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,5 +1,7 @@
 // index.ts - Full extension entry point with commands
 import type { ExtensionAPI, ExtensionContext, ToolInfo } from "@mariozechner/pi-coding-agent";
+import { keyHint } from "@mariozechner/pi-coding-agent";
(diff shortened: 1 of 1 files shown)


titouanmathis:
Sorry for the noise, clanker got ahead, i'll reopen with better commits and descriptions.

PR #12, opened 2026-02-25
feat: improve output for mcp query and result

titouanmathis:
MCP tool calls have no visual feedback for the query being run and output result is displaid in full.

This PR adds support both `renderCall` and `renderResult` to both direct tools and the mcp proxy tool to match behavior from other pi tools (read, write, etc.). The tool name and args are displayed inline on invocation, and results are collapsed to 10 lines with a native expand hint with `ctrl+o`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the renderer work, @titouanmathis. The current shared MCP renderer now covers direct and proxy calls, parsed arguments, compact terminal-width-aware output, expanded results, images, partial output, errors, and theme-less hosts, with focused tests and README documentation. I’m closing this older proposal as superseded by that broader implementation.

PR #14, opened 2026-02-28
fix(mcp-panel): sanitize multiline text and make Keep & Close apply changes

gpmarques:
## What

This PR applies a minimal, targeted fix to `mcp-panel.ts` for two UX/rendering issues in the `/mcp` TUI panel.

### Changes included

1. **Sanitize display text**
   - Added `sanitizeDisplayText(...)` to normalize external metadata used in rendering.
   - Strips ANSI escape sequences and control characters.
   - Collapses newlines/tabs/whitespace into single-line display text.

2. **Keep discard behavior**
   - In discard confirmation, selecting **Keep** on `Enter` now closes the panel and applies changes (`done(buildResult())`) instead of only leaving confirmation mode.

3. **Clarify button label**
   - Renamed confirmation button from **Keep** to **Keep & Close** to match behavior.

4. **Protect row rendering from multiline content**
   - Updated `row(...)` to force one-line content before truncation, preventing box/layout corruption from multiline/control characters.

5. **Use sanitized values in server/tool rows**
   - `renderServerRow(...)`: sanitize `server.name` and `server.importKind`.
   - `renderToolRow(...)`: sanitize `tool.name` and `tool.description`.
   - Width/truncation now use sanitized text, improving stable alignment.

---

## Why

Two user-facing issues were reproducible in the current panel behavior:

1. **Layout breaks with multiline tool descriptions**
   - Some MCP metadata (e.g., descriptions with line breaks/args blocks or control chars) can spill across lines and break bordered row rendering.
   - Sanitizing + enforcing single-line row rendering makes the TUI robust against untrusted/irregular metadata.

2. **“Keep” in discard confirm is misleading**
   - Previously, **Keep** only closed the confirmation prompt, not the panel, which is unexpected and easy to misinterpret.
   - Updating the action to “keep changes and close” plus renaming to **Keep & Close** aligns behavior with user expectation and label semantics.

### DIFF
--- a/mcp-panel.ts
+++ b/mcp-panel.ts
@@ -34,6 +34,16 @@ function fg(code: string, text: string): string {
   return `\x1b[${code}m${text}\x1b[0m`;
 }
 
+function sanitizeDisplayText(text: string): string {
+  return text
+    .replace(/\x1B\[[0-?]*[ -/]*[@-~]/g, "")
+    .replace(/\x1B\][^\x07\x1B]*(?:\x07|\x1B\\)/g, "")
+    .replace(/[\r\n\t]+/g, " ")
+    .replace(/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/g, "")
+    .replace(/\s+/g, " ")
+    .trim();
+}
+
 const RAINBOW_COLORS = [
   "38;2;178;129;214",
   "38;2;215;135;175",
@@ -452,7 +462,8 @@ class McpPanel {
         this.cleanup();
         this.done({ cancelled: true, changes: new Map() });
       } else {
-        this.confirmingDiscard = false;
+        this.cleanup();
+        this.done(this.buildResult());
       }
       return;
     }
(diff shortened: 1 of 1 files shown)


cullenbmacdonald:
i'm hitting this issue with an atlassian mcp server that has a lot of new line characters in the tool descriptions, so it would be cool to get a fix in. this PR seems to do it.

xuli500177:
## Additional Case: Windows Terminal Rendering Issue

I can confirm this issue also affects the `/mcp` panel in **Windows Terminal** (full-screen mode).

### Symptoms
- When pressing arrow keys (↑/↓) to navigate the MCP server/tool list, the entire panel rendering becomes corrupted
- The layout breaks and displays incorrectly until the panel is closed and reopened
- This occurs **only** with the `/mcp` command panel; other PI selectors (like `/model`, `/settings`) work fine

### Environment
- **Terminal**: Windows Terminal (full-screen)
- **OS**: Linux (WSL2)
- **pi-mcp-adapter version**: 2.9.0
- **PI version**: 0.78.1

### Additional Observations
The issue seems related to how the `McpPanel` handles ANSI escape sequences and width calculations during re-renders. The `renderToolRow` method uses hardcoded values (`7` and `8`) for prefix length that may not correctly account for ANSI codes:

```typescript
const prefixLen = 7 + visibleWidth(tool.name);
const maxDescLen = Math.max(0, innerW - prefixLen - 8);
```

This fix (sanitizing multiline text and enforcing single-line rendering) would likely resolve the Windows Terminal issue as well.

Happy to provide more details or test a fix if needed.

PR #25, opened 2026-03-23
feat: add OpenCode config import support

NicoAvanzDev:
## Summary
- add `opencode` to supported `ImportKind` values
- import MCP servers from OpenCode global config (`~/.config/opencode/opencode.json`) and project config (`./opencode.json`)
- map OpenCode local/remote MCP definitions into Pi MCP server entries
- document OpenCode in the README and add coverage for global/project import behavior

## Testing
- npm test
- cd examples/interactive-visualizer && npm run build


21307369:
Isn't it better to directly use the mcp configuration from other tools? Why do we need to import it?

PR #32, opened 2026-04-02
feat: Fix oAuth support using MCP SDK & add good oAuth discovery

jordyvandomselaar:
Heya Nico!

 Thank you for creating this package! I've been running my own fork with some bugfixes that I think would be beneficial to merge back into the original package.

 These changes:

 - make HTTP MCP OAuth work end-to-end using the SDK-backed auth flow
 - store tokens and client registration durably so auth survives restarts
 - improve /mcp auth and /mcp-auth so users can see auth status and retry auth more reliably
 - only trigger browser-based auth on intentional retries; background reconnects stay non-interactive
 - support both authorization_code and client_credentials
 - detect OAuth automatically for HTTP MCP servers that advertise OAuth metadata, even without explicit OAuth config
 - migrate legacy stored OAuth tokens forward automatically

 I also added test coverage around the auth store, command flow, and HTTP transport behavior.

 Happy to split this up if you'd prefer a smaller PR.


nicobailon:
Thanks for the contribution! The package now has working HTTP OAuth via the SDK (v2.3.0+), durable tokens, non-interactive reconnects, and client_credentials support (v2.3.2). The auth flow and panel behavior you described are already implemented in the current architecture, so this PR would create significant conflicts. Closing as the functionality is now in place.

PR #33, opened 2026-04-05
feat: Add automatic OAuth refresh and browser re-auth

mfalkiewicz:
## Summary
- add OAuth browser sign-in flow to `/mcp-auth` using discovery, dynamic client registration, PKCE, and localhost callback
- persist OAuth metadata needed for token refresh
- refresh expired access tokens automatically before connecting
- trigger interactive browser re-auth on session start or lazy connect when no valid OAuth token remains
- add `settings.autoOauthBrowserAuth` (default `true`) so automatic browser popup behavior is configurable

## Docs
- update README to document browser-based OAuth, automatic refresh/re-auth, and the new `autoOauthBrowserAuth` setting
- update changelog `Unreleased` section

## Notes
- browser-based re-auth only runs in interactive sessions
- non-interactive sessions still fail closed with needs-auth behavior
- automatic browser popup is enabled by default, but can be disabled with `settings.autoOauthBrowserAuth: false`

## Tested
- completed Atlassian Rovo MCP OAuth flow locally
- verified token persistence and successful initialize request against `https://mcp.atlassian.com/v1/mcp`
- verified module import after patching with `tsx`
- added focused tests for oauth token persistence/refresh and updated lifecycle mocks for session-start auto auth

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Interactive OAuth browser flow for `/mcp-auth` using discovery, dynamic client registration, PKCE, and a localhost callback.
+- Automatic OAuth access-token refresh using persisted token endpoint/client metadata.
+- Automatic interactive browser re-auth on session start and lazy connect when no valid OAuth token remains.
+- `settings.autoOauthBrowserAuth` to disable automatic browser popups and require manual `/mcp-auth` instead.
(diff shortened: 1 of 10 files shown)


mfalkiewicz:
Closing this since it overlaps heavily with #32, which is broader and follows the SDK-backed OAuth direction.

The main thing this PR added beyond that was automatic browser re-auth by default (with a config toggle to disable it), but #32's intentional/no-surprise-browser policy is probably the better default for upstream.

Thanks — I’d rather not split review effort across two very similar OAuth PRs.

PR #35, opened 2026-04-07
feat(oauth): real OAuth 2.1 flow with dynamic client registration and loopback callback

shuv1337:
## Summary

Replaces the current 'bring-your-own-pre-fetched-token' OAuth path with a real OAuth 2.1 flow that drives the SDK's built-in auth machinery end-to-end: discovery, dynamic client registration (RFC 7591), PKCE, browser-based consent on a loopback callback, code exchange, and automatic refresh.

## Motivation

Today, servers configured with `auth: "oauth"` require the user to manually obtain an access token elsewhere and drop it in `~/.pi/agent/mcp-oauth/<server>/tokens.json`. The README describes this accurately:

> - OAuth tokens obtained externally (no browser flow)
> - No automatic token refresh

The MCP SDK that this adapter already bundles has full `OAuthClientProvider` support — including dynamic client registration, PKCE, auto-refresh, and 401 re-auth — but `createHttpTransport` just never passes an `authProvider` to `StreamableHTTPClientTransport`. That made hosted servers like Supabase, Linear, and GitHub's MCP (all of which default to OAuth 2.1 + dynamic registration) a dead-end:

```
Error: MCP: Failed to connect to supabase: SSE error: Non-200 status code (401)
```

## What this PR does

1. **`oauth-provider.ts`** — `FileBackedOAuthProvider` implementing the SDK's `OAuthClientProvider` interface. Persists tokens / dynamic-registration client info / PKCE code verifier under `~/.pi/agent/mcp-oauth/<server>/` with mode `0600`. **Backwards compatible** with the legacy hand-rolled `tokens.json` format so existing setups keep working.

2. **`oauth-flow.ts`** — two small helpers with zero extra dependencies:
   - `startCallbackServer()` binds a single-shot HTTP server on `127.0.0.1:<random-ephemeral-port>` and exposes `waitForCode()`. Uses a stored-result / pending-listeners design so a request landing before the consumer attaches handlers doesn't trigger spurious 'unhandled rejection' warnings.
   - `openBrowser()` cross-platform browser launcher that honors `PI_MCP_BROWSER=none` for headless / SSH environments.

3. **`server-manager.ts`** — new `createOAuthHttpTransport()` helper. Probe-connects with the auth provider attached; if cached tokens work, returns a fresh transport immediately. On `UnauthorizedError`, waits on the loopback callback, calls `auth(provider, { authorizationCode })` to exchange the code for tokens, invalidates the now-stale PKCE verifier, and returns a fresh transport for the real connection. OAuth servers skip the SSE fallback path.

4. **`commands.ts` + `index.ts`** — `/mcp-auth <server>` no longer just prints instructions. It triggers a reconnect through the server manager, which runs the real OAuth flow end-to-end. The old instruction text is kept as a fallback for the no-state case.

5. **`types.ts`** — new optional `oauthScope` field on `ServerEntry` for explicitly requesting OAuth scope(s).

6. **`package.json`** — adds `oauth-flow.ts` and `oauth-provider.ts` to the published `files` list.

## Config example

```json
{
  "mcpServers": {
    "supabase": {
      "type": "http",
      "url": "https://mcp.supabase.com/mcp",
      "auth": "oauth"
    }
  }
}
```

That's it. On first use, browser opens, user consents, tokens persist, done. On every subsequent use, cached tokens are loaded and refreshed automatically when they expire.

## Tests

**19 new unit tests** (`197 passing` total, up from `178`), plus an end-to-end live smoke test I ran against the real `https://mcp.supabase.com/mcp` during development (not checked in \u2014 it makes network calls). The live test confirmed:

- ✅ Protected-resource metadata discovery (`api.supabase.com`)
- ✅ Dynamic client registration (real `client_id` issued)
- ✅ PKCE S256 challenge generation
- ✅ Loopback callback server binding
- ✅ Browser opened to the authorization URL
- ✅ Code exchange after consent
- ✅ Token persistence with correct `0600` mode
- ✅ Fresh connect with tokens → **29 real Supabase MCP tools listed**

No new type errors; baseline of 11 pre-existing `tsc --noEmit` errors (unrelated to OAuth, all in modules provided by pi at runtime) is unchanged. The only pre-existing test failures (2x `interactive-visualizer` tests needing dist files) are also unchanged.

## Backwards compatibility

- The on-disk format of `tokens.json` is the same as before, and the new provider reads the legacy shape. Users who were dropping hand-rolled tokens into `~/.pi/agent/mcp-oauth/<server>/tokens.json` don't need to change anything.
- Existing `auth: "bearer"` config with `bearerToken` / `bearerTokenEnv` is unchanged.
- Servers that don't set `auth` at all go through the existing probe-and-SSE-fallback path unchanged.
- `oauth-handler.ts` is kept for compatibility with the mcp-panel `getConnectionStatus` path that reads stored tokens directly.

## Env vars

- `PI_MCP_BROWSER=none` — disable auto browser open. The user can copy the 'OAuth: open this URL to authorize' line from the log into a browser on another machine (use SSH port forwarding for the loopback callback if needed).
- `PI_MCP_BROWSER=<command>` — override the opener binary.


shuv1337:
Closing — opened by mistake without the owner's request. The branch remains on my fork at shuv1337/pi-mcp-adapter:feat/oauth-real-flow if useful later.

PR #37, opened 2026-04-09
feat: add excludeTools option to blacklist tools per server

### LINKED ISSUE #36: Feature: Add excludeTools option to blacklist specific tools per server
## Problem

Currently, `directTools` supports `true` (all tools) or a `string[]` allowlist. There's no way to exclude specific tools while keeping all others.

This is a problem when:
- An MCP server exposes tools you never want (e.g., Code Connect or FigJam tools from Figma MCP)
- You want new tools added by the server to automatically appear without updating your config
- Maintaining an allowlist is tedious when you only want to exclude a few tools out of many

## Proposed Solution

Add an `excludeTools` option to the server config that accepts an array of tool names to hide:

```json
{
  "mcpServers": {
    "figma": {
      "url": "http://localhost:3845/mcp",
      "directTools": true,
      "excludeTools": [
        "get_code_connect_map",
        "add_code_connect_map",
        "get_code_connect_suggestions",
        "send_code_connect_mappings",
        "get_figjam"
      ]
    }
  }
}
```

This should filter tools from:
- Direct tool registration
- Proxy search/list/describe results
- The `/mcp` panel UI

The option should match against both original MCP tool names and prefixed names (e.g., both `get_screenshot` and `figma_get_screenshot` should work).

## Use Case

Figma's MCP server exposes 10 tools. I never use Code Connect or FigJam (5 tools). With `directTools: true` + `excludeTools`, I get all the tools I want today, and any new tools Figma adds in the future will automatically appear — without touching my config.

ahmadaccino:
Closes #36

## Summary

Adds an `excludeTools` server config option that accepts an array of tool names to exclude. This complements `directTools` by allowing a blacklist approach: register all tools except specific ones, rather than maintaining an explicit allowlist.

## Changes

| File | Change |
|------|--------|
| `types.ts` | Added `excludeTools?: string[]` to `ServerEntry` |
| `tool-metadata.ts` | `buildToolMetadata` skips excluded tools (live connections) |
| `metadata-cache.ts` | `reconstructToolMetadata` accepts + applies `excludeTools` (cached metadata) |
| `init.ts` | Passes `definition.excludeTools` to `reconstructToolMetadata` |
| `direct-tools.ts` | Skips excluded tools when registering direct tools |
| `mcp-panel.ts` | Skips excluded tools in the `/mcp` panel UI |

## Behavior

- Matches against both original MCP names (`get_screenshot`) and prefixed names (`figma_get_screenshot`), so either format works
- Excluded tools are filtered from direct registration, proxy search/list/describe, and the `/mcp` panel
- Excluded tools cannot be called via `mcp({ tool: ... })` either, since they're removed from the metadata map
- Works with `directTools: true`, `directTools: ["..."] `, or proxy-only mode

## Example Config

```json
{
  "mcpServers": {
    "figma": {
      "url": "http://localhost:3845/mcp",
      "directTools": true,
      "excludeTools": ["get_code_connect_map", "get_figjam"]
    }
  }
}
```

### DIFF
--- a/direct-tools.ts
+++ b/direct-tools.ts
@@ -66,7 +66,10 @@ export function resolveDirectTools(
 
     if (!toolFilter) continue;
 
+    const excludeSet = definition.excludeTools ? new Set(definition.excludeTools) : null;
+
     for (const tool of serverCache.tools ?? []) {
+      if (excludeSet && (excludeSet.has(tool.name) || excludeSet.has(formatToolName(tool.name, serverName, prefix)))) continue;
       if (toolFilter !== true && !toolFilter.includes(tool.name)) continue;
       const prefixedName = formatToolName(tool.name, serverName, prefix);
       if (BUILTIN_NAMES.has(prefixedName)) {
@@ -92,6 +95,7 @@ export function resolveDirectTools(
     if (definition.exposeResources !== false) {
       for (const resource of serverCache.resources ?? []) {
         const baseName = `get_${resourceNameToToolName(resource.name)}`;
+        if (excludeSet && (excludeSet.has(baseName) || excludeSet.has(formatToolName(baseName, serverName, prefix)))) continue;
         if (toolFilter !== true && !toolFilter.includes(baseName)) continue;
         const prefixedName = formatToolName(baseName, serverName, prefix);
         if (BUILTIN_NAMES.has(prefixedName)) {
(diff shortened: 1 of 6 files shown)


nicobailon:
@ahmadaccino thanks, this has been addressed in the latest release, and I've credited you.

PR #39, opened 2026-04-09
docs: Update README with existing 'headers' option

vuon9:
## Changes
Added 'headers' option for HTTP endpoint configuration and http remote example which uses headers.

## Why
I've been studying code to understand it, hoping this addition will be beneficial to others.


nicobailon:
Thanks for the docs cleanup. The README now documents the headers option in the server settings table, including env interpolation, so this doc change is covered on main. Closing this one as already addressed.

PR #43, opened 2026-04-12
fix: prevent sub-agent hang after OAuth callback server is started

blai:
OAuth changes added initializeOAuth() on session_start, which starts an HTTP callback server. Two defects caused sub-agent processes to hang indefinitely after completing their task:

1. session_shutdown did not call shutdownOAuth(), so the callback server was never closed when a session ended. Fix: call shutdownOAuth() alongside shutdownState() in the session_shutdown handler.

2. The HTTP server was created without server.unref(), making it a strong event- loop reference. Even if shutdown ran cleanly, any race where the handler was skipped (e.g. SIGKILL, uncaught throw) would keep the process alive. Fix: call server.unref() immediately after bind so Node treats the server as a background resource and exits naturally when no foreground work remains.

Tests: added regression coverage in two new suites:
- index-lifecycle: asserts shutdownOAuth is called on session_shutdown
- mcp-callback-server-unref: asserts server.unref() is called on successful bind and NOT called on a failed bind (EADDRINUSE)

### DIFF
--- a/__tests__/index-lifecycle.test.ts
+++ b/__tests__/index-lifecycle.test.ts
@@ -5,6 +5,7 @@ const mocks = vi.hoisted(() => ({
   updateStatusBar: vi.fn(),
   flushMetadataCache: vi.fn(),
   initializeOAuth: vi.fn().mockResolvedValue(undefined),
+  shutdownOAuth: vi.fn().mockResolvedValue(undefined),
   loadMcpConfig: vi.fn(() => ({ mcpServers: {} })),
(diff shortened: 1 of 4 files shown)


monochromatti:
I discovered this from subagents hanging indefinitely. Closed my duplicate PR, giving my vote to this, @nicobailon . pi-subagents + pi-mcp-adapter is broken until this is merged.

nicobailon:
@blai @monochromatti thanks, I ended up implementing a broader fix directly based on this.

PR #44, opened 2026-04-13
fix: shut down OAuth callback server on session end

monochromatti:
The OAuth callback HTTP server was started on every session_start but never stopped on session_shutdown, keeping the Node process alive indefinitely. This caused non-interactive pi runs to hang after completing their output. Discovered this because subagents (from pi-subagents) would hang forever.


monochromatti:
Okay I see this is a duplicate of https://github.com/nicobailon/pi-mcp-adapter/pull/43 . Prefer their implementation.

PR #56, opened 2026-04-25
 feat: add support for extension-contributed MCP server definitions

titouanmathis:
This PR adds support for **extension-contributed MCP server definitions**.

Extensions can now register MCP servers programmatically through the adapter instead of requiring users to manually copy server entries into their Pi config first. The adapter resolves those contributed servers together with imported, user, and project config, while preserving the existing override behavior.

This makes it easier for other Pi extensions to ship MCP-backed functionality without forcing users to hand-edit their config before anything works. It can be useful in org-shared Pi extensions for default MCP server configurations. 

It keeps the existing config model intact:

- extensions can provide sensible defaults
- users can still override them
- project config still takes precedence when needed

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,61 @@ Supported compatibility imports: `cursor`, `claude-code`, `claude-desktop`, `vsc
 
 Prefer `.mcp.json` for project-local shared MCP config. Use `.pi/mcp.json` only when you need a Pi-specific project override. Project files override both user-global shared MCP config and Pi global overrides.
 
+## Extension-Contributed Servers
+
+Extensions can contribute MCP server definitions programmatically. This is useful when an extension wants to ship MCP-backed functionality without asking users to manually copy server entries into `~/.pi/agent/mcp.json` first.
+
+Use `registerMcpServerProvider(...)` from `pi-mcp-adapter/providers`:
+
+```ts
+import { registerMcpServerProvider } from "pi-mcp-adapter/providers";
+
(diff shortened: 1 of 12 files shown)


carlos-algms:
I would love to have that, so I can split the MCP list into 2

- work
- personal

I write my own extension that detects the CWD and returns a list of MCP servers: 

```ts
import { registerMcpServerProvider } from "pi-mcp-adapter/providers";
import { homedir } from "node:os";

export default function scopedMcp(pi) {
  registerMcpServerProvider(pi, () => {
    const cwd = process.cwd();
    const home = homedir();
    if (cwd.startsWith(`${home}/work`)) {
      return { source: "scoped-mcp-work", servers: { /* work mcps */ } };
    }
    if (cwd.startsWith(`${home}/projects`)) {
      return { source: "scoped-mcp-personal", servers: { /* personal mcps */ } };
    }
  });
}
```

nicobailon:
Thanks for putting this together, Titouan. I’m closing this branch rather than merging it as-is because the adapter has since moved to the owner/generation lifecycle from #170 and prompt discovery from #203, and this PR rewrites those same config/lifecycle surfaces from an older base.

The useful idea here is extension-provided MCP definitions, but the global merged-config model competes with the isolated programmatic source model in #191. I’m choosing not to ship both public composition APIs. If we later add a global provider layer, I’ll treat this PR as design/reference material and credit you for it in the changelog or release notes.

PR #60, opened 2026-04-27
use 127.0.0.1 loopback IP literal instead of localhost for OAuth flow

mcphailtom:
## Summary

Use the loopback IP literal `127.0.0.1` for the OAuth callback `redirect_uri` (and the matching local listener bind address) instead of the hostname `localhost`.

I hit this while trying to complete the auth flow with the Datadog MCP server (see notes below).

## Why

Per [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3), OAuth native-app clients performing the loopback redirect flow **SHOULD** use the loopback IP literals (`127.0.0.1`, `[::1]`) rather than the hostname `localhost`:

1. `localhost` is resolved through the OS resolver — `/etc/hosts`, NSS, mDNS, sometimes DNS — and can map to `::1`, `127.0.0.1`, or something else depending on host configuration. The IP literals are guaranteed loopback by the IP stack itself.
2. Because of (1), authorization servers can perform an exact-string comparison on the registered `redirect_uri` and know it will actually hit a loopback interface. With `localhost`, they have to either trust the resolver or do fuzzy matching.

Some providers enforce this strictly, e.g. **Datadog's MCP server** (`https://mcp.datadoghq.com/api/unstable/mcp-server/mcp`) accepts the Dynamic Client Registration call (returns a `client_id`) but then rejects the authorize step with "invalid callback URL" because the registered redirect URI contains the literal string `localhost`. Switching to `127.0.0.1` makes the flow succeed without any other change.

## Changes

Two one-line changes:

- **`mcp-oauth-provider.ts`** — `redirectUrl` getter now returns `http://127.0.0.1:<port>/callback`.
- **`mcp-callback-server.ts`** — local HTTP listener binds to `"127.0.0.1"` instead of `"localhost"`, so the redirect always lands on a listening socket regardless of how `localhost` resolves on the host (otherwise, on a system where `localhost` resolves to `::1` first, the browser would hit `127.0.0.1:<port>` and get connection refused).

## Test plan

Verified end-to-end with Datadog's MCP server:
1. Patch applied to globally-installed `pi-mcp-adapter@2.5.1`.
2. Removed cached client registration at `~/.pi/agent/mcp-oauth/datadog/`.
3. Restarted pi, triggered Datadog auth — DCR + authorize + token exchange all succeed; tools list resolves.

Existing flows against providers that accept `localhost` (e.g. GitHub Copilot MCP, Context7) are unaffected since `127.0.0.1` is a strict subset of what those servers accept.


hydrosquall:
Hey @mcphailtom thanks for looking into this 🙇 . I work with the Datadog MCP server! You can reach us on the shared slack if you want to discuss any feedback with the team.

To test a related MCP server change https://github.com/nicobailon/pi-mcp-adapter/pull/65 , I used API-key based authentication . Perhaps this can can unblock folks in the meantime as an alternative to OAuth based login.

mcphailtom:
Superseded by #53 / daae745, though can I just note that fix uses `localhost` rather than the loopback IP literal, which isn't compliant with RFC 8252 §7.3.

But I acknowledge it probably works fine in 99.99999% of cases 😃

PR #62, opened 2026-04-27
feat: Add configurable MCP tool timeout

danecando:
## Summary
- add global and per-server `toolTimeoutMs` configuration support
- pass the resolved timeout through direct-tool, proxy-tool, and UI-initiated MCP tool calls
- document the new setting and add a regression test for the UI server path

## Why
Some MCP tools legitimately run longer than the SDK default timeout. The adapter did not expose a way to tune request timeouts, so those tools could fail even when the server was behaving correctly.

## Impact
Users can now configure `toolTimeoutMs` globally under `settings` or override it per server under `mcpServers.<name>`. Per-server values take precedence over the global default.

## Root Cause
The adapter always called `client.callTool(...)` without a timeout override, so every tool execution path was locked to the SDK default request timeout.

## Validation
- `npx vitest run __tests__/ui-server.test.ts __tests__/proxy-modes-auto-auth.test.ts __tests__/direct-tools-auto-auth.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added global and per-server `toolTimeoutMs` settings to configure MCP tool call request timeouts.
+
 ## [2.6.0] - 2026-05-10
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -110,7 +110,8 @@ Pi-specific files are the write targets for imported or shared global servers wh
       "command": "npx",
       "args": ["-y", "some-mcp-server"],
       "lifecycle": "lazy",
-      "idleTimeout": 10
+      "idleTimeout": 10,
+      "toolTimeoutMs": 600000
     }
   }
 }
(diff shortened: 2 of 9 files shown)


esinecan:
Tested this locally against pi 0.73.0 + a slow MCP server (Python/NotebookLM-backed `search_memory` that routinely takes 30-90s on cold start, occasionally more). With `toolTimeoutMs: 180000` per-server, the call goes through fine; without it the SDK 60s default kills it. Wiring is correct on the callTool path.

Two adjacent gaps I hit while validating, raising here in case they belong on this PR:

1. `readResource` is not threaded with the same timeout. Currently `connection.client.readResource(...)` calls in `direct-tools.ts`, `proxy-modes.ts`, and `server-manager.ts` use the SDK default. For servers that expose UI resources or large blobs, slow reads die at 60s.
2. Startup path (`client.connect`, initial `listTools`, initial `listResources`) is not bounded. If a server hangs at handshake, pi waits indefinitely. A separate `startupTimeoutMs` (per-server + global) would let the adapter mark the server failed and continue.

Happy to send a follow-up PR for either or both, or fold them in here, whichever you prefer. Either way, +1 to merging this as-is.

nicobailon:
Thanks for pushing on configurable MCP timeouts, Dane. Current `main` now ships the broader `requestTimeoutMs` contract across connect, discovery, tool, resource, and UI proxy requests while preserving abort signals and session recovery, so a separate narrower `toolTimeoutMs` would duplicate configuration and create unclear precedence. I added your #62 credit to the shipped timeout changelog entry before closing this as superseded.

PR #64, opened 2026-04-28
feat: add renderCall/renderResult to MCP direct tools

cad0p:
Implements custom TUI rendering for MCP direct tools so they support Ctrl+O collapse/expand in Pi's interactive mode.

**Changes:**
- **renderCall**: compact summary when collapsed (extracts objective/query/url, truncates to 60 chars), pretty-printed JSON when expanded
- **renderResult**: 5-line limit when collapsed, full output when expanded, handles text/image/generic blocks, normalizes \r\n, truncates long lines to 500 chars
- Adds @mariozechner/pi-tui as peerDependency

**Why:**
Without custom renderers, MCP direct tools fall back to plain text display that ignores the \expanded\ state, making Ctrl+O non-functional for these tools.

**Testing:**
Tested locally with \parallel_search_web_search\ — collapse/expand works as expected.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -52,14 +52,19 @@ export default function mcpAdapter(pi: ExtensionAPI) {
   const prefix = earlyConfig.settings?.toolPrefix ?? "server";
 
   const envRaw = process.env.MCP_DIRECT_TOOLS;
-  const directSpecs = envRaw === "__none__"
-    ? []
-    : resolveDirectTools(
-        earlyConfig,
-        earlyCache,
-        prefix,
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for the renderer and direct-tool investigation, @cad0p. The renderer portion is superseded by the current shared direct/proxy renderer and tests. The remaining commit makes explicit MCP_DIRECT_TOOLS=__none__ fall back to configured direct tools; current contract tests preserve __none__ as a hard disable, while configured cached tools already register when the variable is absent. I’m closing this stale branch rather than merging those semantics.

PR #65, opened 2026-04-29
fix(mcp-ui): Fix CSP construction to match the fields used in the MCP Apps spec

hydrosquall:
## Summary

There was a mismatch between the pi internal `UiResourceCsp` type and the [MCP Apps spec](https://raw.githubusercontent.com/modelcontextprotocol/ext-apps/refs/heads/main/specification/2026-01-26/apps.mdx) (see the Content Security Policy Enforcement section). 

This meant CSP metadata from a spec-compliant server was being silently ignored. As a result, some MCP apps that was working in [Cursor](https://cursor.com/blog/new-plugins#infrastructure) did not load in Pi-mcp-adapter because the resourceDomains field was not being checked.

## Changes

1. The `UiResourceCsp` type had granular fields (`scriptDomains`, `styleDomains`, `fontDomains`, `imgDomains`, `mediaDomains`, `workerDomains`) that don't exist in the current version of the [spec](https://raw.githubusercontent.com/modelcontextprotocol/ext-apps/refs/heads/main/specification/2026-01-26/apps.mdx) (maybe these were from previous versions?)

The spec uses a single `resourceDomains` field that maps to all static-asset directives, so those have been replaced.

2. `buildCspMetaContent` was also missing some required base values from each directive (`'self' 'unsafe-inline'` on script/style, `data:` on img/media, `object-src 'none'`), and `frame-src` and `base-uri` were only emitted when extra domains were provided. The spec requires them unconditionally with safe fallbacks.

2. Finally, when `csp` was omitted entirely, the function returned `undefined` so nothing got injected. The spec mandates a restrictive default in that case, so the function now always returns a string.

Files changed:

- `types.ts`: replace the granular domain fields with spec-defined `resourceDomains`, `connectDomains`, `frameDomains`, `baseUriDomains`
- `host-html-template.ts`: rewrite `buildCspMetaContent` to always return a string with correct base values; tighten `applyCspMeta` signature to match
- tests updated to use the new field names and verify the spec-correct directive output

## Testing

Existing vitest suite covers the updated field names and the new directive assertions. Run with `npm test`.

With this change in place, when interacting with the Datadog MCP [server](https://docs.datadoghq.com/bits_ai/mcp_server/), you'll be able to see graphs (instead of a blank screen). This is using fixture data inside the Glimpse viewer.

https://github.com/user-attachments/assets/875b0b68-adee-4c9b-a2e4-2ea11daa1615

## Notes

- If there are some other apps that are dependent on those other paths, this change might break them. However, I couldn't find any examples in this repo. Given that mcp-ui is now aligned with mcp-apps, I figured this change could be helpful to share with the community rather than forking.

### DIFF
--- a/host-html-template.ts
+++ b/host-html-template.ts
@@ -356,42 +356,48 @@ export function buildHostHtmlTemplate(input: HostHtmlTemplateInput): string {
 </html>`;
 }
 
-export function buildCspMetaContent(csp: UiResourceCsp | undefined): string | undefined {
-  if (!csp) return undefined;
+/**
+ * Build a CSP string from resource metadata following the MCP Apps spec
+ * (specification/2026-01-26/apps.mdx § "Content Security Policy Enforcement").
+ *
+ * When `csp` is omitted the host MUST apply the spec-mandated restrictive default.
+ * When `csp` is provided the host builds a full directive list using `resourceDomains`
+ * for static assets and `connectDomains` for network access.
+ */
+export function buildCspMetaContent(csp: UiResourceCsp | undefined): string {
+  // Spec: "If ui.csp is omitted, Host MUST use" this restrictive default.
+  if (!csp) {
+    return [
+      "default-src 'none'",
+      "script-src 'self' 'unsafe-inline'",
+      "style-src 'self' 'unsafe-inline'",
+      "img-src 'self' data:",
+      "media-src 'self' data:",
+      "connect-src 'none'",
+      "object-src 'none'",
+    ].join("; ");
+  }
 
-  const directives: string[] = [];
-  directives.push("default-src 'none'");
-
-  const scriptSrc = toDirective("script-src", csp.scriptDomains);
-  const styleSrc = toDirective("style-src", csp.styleDomains);
-  const fontSrc = toDirective("font-src", csp.fontDomains);
-  const imgSrc = toDirective("img-src", csp.imgDomains);
-  const mediaSrc = toDirective("media-src", csp.mediaDomains);
-  const connectSrc = toDirective("connect-src", csp.connectDomains);
-  const frameSrc = toDirective("frame-src", csp.frameDomains);
-  const workerSrc = toDirective("worker-src", csp.workerDomains);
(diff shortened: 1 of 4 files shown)


host-html-template.ts
```diff
@@ -356,42 +356,48 @@ export function buildHostHtmlTemplate(input: HostHtmlTemplateInput): string {
...
 
-function toDirective(name: string, domains: string[] | undefined): string | null {
-  if (!domains || domains.length === 0) return null;
-  return `${name} ${domains.join(" ")}`;
+  // Spec CSP construction from declared domains.
+  const res = csp.resourceDomains?.join(" ") ?? "";
+  const connect = csp.connectDomains?.join(" ") ?? "";
+  const frame = csp.frameDomains?.join(" ") || "'none'";
+  const base = csp.baseUriDomains?.join(" ") || "'self'";
+  return [
```
hydrosquall, line 386: Reference in the official spec https://raw.githubusercontent.com/modelcontextprotocol/ext-apps/refs/heads/main/examples/basic-host/serve.ts

hydrosquall:
Hi @nicobailon , I was curious if you had a take on whether the direction of PR might be a good fit for the main project, as I have a few more stacked PRs on top that would combine to adjust the adapter for compatibility with production MCP apps

If it's not a good fit, I can close the PR and just keep the changes on a fork for our own testing (but I would rather share the change with the community if it does fit). 

Thanks for your consideration and this great contribution to the Pi ecosystem 🙇

nicobailon:
Thanks for sending this over. The CSP field alignment was the right direction and helped catch the spec mismatch. I carried the change forward on an upstream branch with a few additional hardening fixes around host-owned CSP enforcement, response headers, stale CSP meta replacement, and CSP metadata validation, so I’m closing this PR in favor of that branch.

hydrosquall:
Thanks @nicobailon !

For my learning, which is the upstream branch name? I have a few additional
updates I can submit (including handling for responsive sizing) which could
use that upstream branch as the base

PR #66, opened 2026-05-01
Localize MCP setup and auth status strings

jerryfan:
I found one small UX issue: the setup/auth path has dense status strings exactly where users are deciding which MCP configs to import, when to authenticate, and when to reconnect. If those messages are misread, Pi can stay disconnected or load the wrong MCP tools into context.

This PR makes that path optionally localizable while preserving the current English strings as fallbacks.

- No new dependency
- No behavior change without an i18n provider
- English remains the default/fallback
- Locales: ja, zh-CN, de — large developer markets for dense setup/error copy
- Small diff; easy to revert

Validation:
- npm pack --dry-run
- npx vitest run __tests__/index-lifecycle.test.ts __tests__/package-manifest.test.ts

Note: full npm test has existing environment-related failures in config/fixture tests on this Windows machine, plus a missing optional @mariozechner/pi-tui package for one panel test. The targeted lifecycle/package checks pass.

### DIFF
--- a/commands.ts
+++ b/commands.ts
@@ -19,42 +19,45 @@ import { supportsOAuth, authenticate } from "./mcp-auth-flow.js";
 import { hasStoredTokens } from "./mcp-auth.js";
 import { loadOnboardingState, markSetupCompleted as persistSetupCompleted, markSharedConfigHintShown } from "./onboarding-state.js";
 import { openPath } from "./utils.js";
+import { t } from "./i18n.js";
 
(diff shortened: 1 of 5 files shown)


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

PR #67, opened 2026-05-01
Localize MCP status empty-state copy

jerryfan:
Reposting this as a much smaller follow-up after retracting the earlier over-broad PR.

The `/mcp` status empty-state is a setup decision point: users need to know whether servers are connected, cached, failed, need auth, or are not configured yet. I kept this PR limited to that status/empty-state path instead of wrapping the setup panel or broader MCP copy.

What changed:
- added a tiny optional i18n bridge in `i18n.ts`
- registered `es`, `fr`, and `pt-BR` strings
- wrapped only `/mcp` server status labels and the no-server setup hint
- added `i18n.ts` to package files so the extension still packs correctly

What did not change:
- no required dependency
- no behavior change without a Pi i18n provider
- English fallback strings stay in the call sites
- no README, setup panel, auth flow, or tool behavior changed
- easy to revert by deleting `i18n.ts` and the small wrappers

Validation:
- `npx esbuild index.ts --bundle --platform=node --format=esm --outfile=... --external:@mariozechner/pi-coding-agent --external:@mariozechner/pi-tui` passed
- `npm pack --dry-run` passed and includes `i18n.ts`
- `npm test` ran: 224 passed / 15 failed; failures appear pre-existing/environmental around Windows path/home config, missing interactive-visualizer dist artifacts, and onboarding state leakage, not these status string wrappers

If useful later, I’m happy to handle broader localization in small follow-up PRs using whatever locale set you prefer.

### DIFF
--- a/commands.ts
+++ b/commands.ts
@@ -19,42 +19,48 @@ import { supportsOAuth, authenticate } from "./mcp-auth-flow.js";
 import { hasStoredTokens } from "./mcp-auth.js";
 import { loadOnboardingState, markSetupCompleted as persistSetupCompleted, markSharedConfigHintShown } from "./onboarding-state.js";
 import { openPath } from "./utils.js";
+import { t } from "./i18n.js";
 
 export async function showStatus(state: McpExtensionState, ctx: ExtensionContext): Promise<void> {
   if (!ctx.hasUI) return;
 
-  const lines: string[] = ["MCP Server Status:", ""];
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks for exploring localized MCP status copy. The current Pi host still has no documented i18n extension API or consumer for the `pi-core/i18n/*` event channels, so this bridge would not run as submitted. The branch is also far behind the current host package names and has no production-path locale tests. I’m closing this rather than leaving it blocked indefinitely; this can be revisited once Pi exposes a supported localization API.

PR #72, opened 2026-05-02
Hot-load refreshed direct MCP tools

devinbost:
## Summary
- hot-load direct MCP tools after cache metadata refreshes instead of requiring a Pi reload
- refresh direct-tool registration after reconnects, lazy/proxy reconnects, and /mcp direct-tool panel changes
- deactivate stale direct tools from the active tool set when refreshed metadata/config no longer exposes them
- update docs/changelog and add lifecycle coverage for hot-load and stale-tool deactivation

## Verification
- npm test
- npm test -- __tests__/index-lifecycle.test.ts __tests__/commands-onboarding.test.ts __tests__/proxy-modes-auto-auth.test.ts
- git diff --check

Note: raw npx tsc --noEmit still reports existing project type issues around host Pi package types/OAuth fixtures/sampling auth types; the Vitest suite is green after building the interactive visualizer fixture.


devinbost:
Updated this branch after opening the companion Pi core PR: https://github.com/badlogic/pi-mono/pull/4084

The adapter now uses optional pi.unregisterTool() when present so stale hot-loaded direct tools are removed from the Pi extension registry, while preserving the previous setActiveTools() fallback for older Pi versions. Full adapter Vitest passes locally: 30 files, 248 tests.

devinbost:
Final update: I also wired MCP server list-change notifications into the same hot-load path. Servers advertising tools.listChanged or resources.listChanged now refresh cached metadata and update direct tools in the current session. Full adapter Vitest passes locally: 30 files, 249 tests. The companion Pi core branch remains at https://github.com/badlogic/pi-mono/pull/4084, which is auto-closed by the upstream contributor gate until a contribution issue gets maintainer lgtm.

esinecan:
Tested 2.5.4 vs this branch (HEAD `dd10133`) against a progressive-disclosure MCP server (faro-mcp v2, advertises `tools.listChanged: true`). Same prompt, same model, same `mcp.json`, only the adapter swapped between runs. Three calls through the unified `mcp` proxy:

1. `discover_tools` with `query: "activity"` (lists a disabled tool)
2. `discover_tools` with `enable: ["feed_get_feed"]` (server confirms enabled)
3. `feed_get_feed` with arbitrary args

On 2.5.4: step 3 returns "Tool not found" and the proxy lists only the 24 tools cached at connect.

On this branch: step 3 reaches the server and returns a normal MCP `-32602` (missing required params), confirming the tool got registered after the enable.

Works for me :+1:

PR #77, opened 2026-05-02
feat: add Pi session id to MCP tool call metadata

barisgit:
## Summary
- Add `pi/session_id` to downstream MCP `tools/call` request metadata
- Preserve existing request `_meta` values while ensuring Pi owns the session id field
- Propagate the same metadata through proxy, direct tool, and MCP UI bridge tool-call paths

## Why
Stateful MCP servers need a stable per-Pi-session identity to route per-conversation state without relying on environment-variable hacks. Pi already has a persistent session id via `ctx.sessionManager.getSessionId()`, so this exposes it per call under Pi's own `_meta` namespace.

## Tests
- `bunx vitest run __tests__/state.test.ts __tests__/direct-tools-auto-auth.test.ts __tests__/proxy-modes-auto-auth.test.ts __tests__/ui-server.test.ts`

### DIFF
--- a/__tests__/direct-tools-auto-auth.test.ts
+++ b/__tests__/direct-tools-auto-auth.test.ts
@@ -66,6 +66,7 @@ describe("direct tools auto auth", () => {
       failureTracker: new Map(),
       ui: { setStatus: vi.fn() },
       completedUiSessions: [],
+      sessionId: "pi-session-direct",
     } as any;
 
     const executor = createDirectToolExecutor(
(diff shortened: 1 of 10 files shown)


barisgit:
Whoops. I said to my pi that let's create a new branch for this and I will later open a PR to this repo, but I guess that it opened the PR as well. Either way this is something that is useful for many mcp servers but I'm not sure everyone would want to gave this on all the time. So should I change something about this feature or maybe gate it behind config?

nicobailon:
Thanks for exploring session metadata propagation. I’m not comfortable sending a persistent Pi session identifier to every configured MCP server by default, especially remote third-party servers, and this PR has no opt-in or allowlist boundary. There is not a concrete product requirement here that justifies adding that disclosure surface. I’m closing this version; a future proposal would need an explicit opt-in use case, privacy documentation, and current-main direct/proxy/UI coverage.

PR #81, opened 2026-05-04
feat: respect modelPreferences.hints in MCP sampling

### LINKED ISSUE #80: feat: respect modelPreferences.hints in MCP sampling
## Problem

When an MCP server sends `sampling/createMessage` with `modelPreferences.hints`, the adapter ignores them and always uses the current conversation model.

This means servers can't request cheaper/faster models for bulk operations. For example, an i18n MCP server translating to 16 locales ends up using Claude Opus for every request — slow, expensive, and hits rate limits.

## MCP spec reference

[`modelPreferences.hints`](https://modelcontextprotocol.io/docs/concepts/sampling#model-preferences) is an ordered list of `ModelHint` objects. Each hint has a `name` field that should be treated as a **case-insensitive substring** match against available model names.

## Current behavior

`resolveSamplingModel()` picks `getCurrentModel()` first, then falls back to any available model — hints are never consulted.

## Proposed fix

Before falling back to the current model, iterate through `hints` and fuzzy-match each against `modelRegistry.getAvailable()` (checking `provider/id` and `model.name`). First match with valid auth wins.

Related: #52

fabkho:
Closes #80

## What

When an MCP server sends `sampling/createMessage` with `modelPreferences.hints`, the adapter now tries to match those hints against available Pi models before falling back to the current conversation model.

## Why

MCP servers like [the-i18n-mcp](https://github.com/fabkho/the-i18n-kit) use sampling for bulk operations (translating to 16+ locales). They send hints like `["flash", "haiku", "gpt-4o-mini"]` requesting fast/cheap models, but the adapter was ignoring them — always using the conversation model (e.g. Claude Opus), which is slow, expensive, and hits rate limits.

## How

In `resolveSamplingModel()`:
1. If `params.modelPreferences.hints` exists, iterate through hints
2. For each hint, fuzzy-match `hint.name` (case-insensitive substring) against `provider/id` and `model.name` of all available models
3. First match with valid auth is used
4. If no hint matches, fall back to existing behavior (current model → any available)

This follows the [MCP spec](https://modelcontextprotocol.io/docs/concepts/sampling#model-preferences) which says hints should be treated as substrings of model names.

## Tests

Added 2 tests:
- ✅ `respects modelPreferences.hints to select a matching model` — verifies flash model is selected when hint matches
- ✅ `falls back to current model when hints don't match` — verifies graceful fallback

All existing tests continue to pass.

### DIFF
--- a/__tests__/sampling-handler.test.ts
+++ b/__tests__/sampling-handler.test.ts
@@ -130,6 +130,74 @@ describe("sampling handler", () => {
     expect(ui.confirm.mock.calls[1][1]).toContain("Bonjour");
   });
 
+  it("respects modelPreferences.hints to select a matching model", async () => {
+    const { handleSamplingRequest } = await import("../sampling-handler.ts");
+    const flashModel = {
+      provider: "google",
+      id: "gemini-2.5-flash",
+      api: "google-genai",
+      name: "Gemini 2.5 Flash",
+      input: ["text"],
+      reasoning: false,
+      cost: usage.cost,
+      contextWindow: 1000000,
+      maxTokens: 8192,
+    };
+    const options = createOptions({
+      getCurrentModel: vi.fn(() => model),
+      modelRegistry: {
+        getAvailable: vi.fn(() => [model, flashModel]),
+        getApiKeyAndHeaders: vi.fn(async () => ({ ok: true, apiKey: "gkey", headers: {} })),
+      },
+    });
+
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the PR. The sampling code on main now honors MCP modelPreferences.hints before falling back to the current/default model, so this is already covered and issue #80 is closed. I’m going to close this stale branch.

PR #83, opened 2026-05-05
Incentivize LLM to describe tools before using them

momo5502:
I have noticed a lot of tool call failures when using this plugin.
It seems that LLMs don't make use of "describe" and instead try to guess schemas.

This change tries to incentivize LLMs to describe tools first, instead of blindly guessing, wasting tokens or using tools in suboptimal ways.


momo5502:
i just saw that the line above my mandatory rule incentivies the LLM to call tools first, which is mutually exclusive to my addition. i will try to see if there is a chance that the mcp servers I work with can be improved so that guessing yields a higher chance of success.

momo5502:
if that doesn't work, i still suggest changing the order of describe and tool calls

PR #89, opened 2026-05-07
Add per-server OAuth callback port

JFanDev:
## Summary

Adds an optional per-server `oauth.callbackPort` setting for OAuth MCP servers.

This lets pre-registered OAuth clients use the exact localhost callback port registered with their OAuth provider without changing `MCP_OAUTH_CALLBACK_PORT` for every MCP server in the Pi process.

Example:

```json
{
  "mcpServers": {
    "slack": {
      "url": "https://mcp.slack.com/mcp",
      "auth": "oauth",
      "oauth": {
        "clientId": "1601185624273.8899143856786",
        "callbackPort": 3118
      }
    }
  }
}
```

## Motivation

Some OAuth servers use pre-registered public clients and require an exact redirect URI. Today the adapter only supports a process-wide callback port via `MCP_OAUTH_CALLBACK_PORT`, so one server that requires a specific port forces all OAuth servers in the same Pi process to use that port.

A per-server option avoids breaking other OAuth flows while still supporting exact-redirect providers.

## Changes

- Adds `oauth.callbackPort` to config types.
- Passes the configured port into OAuth callback server startup.
- Keeps existing default behavior when `callbackPort` is omitted.
- Documents the new option in README.
- Adds callback server coverage for preferred port binding.

## Test plan

- `node --import tsx --test mcp-callback-server.test.ts mcp-auth-flow.test.ts mcp-oauth-provider.test.ts`

Also ran `npm test`; unrelated existing failures occurred due missing generated interactive visualizer dist files and missing `@mariozechner/pi-tui` package in a fresh clone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ Pi-specific files are the write targets for imported or shared global servers wh
 | `headers` | HTTP headers; supports `${VAR}` and `$env:VAR` interpolation |
 | `auth` | `"bearer"` or `"oauth"` |
 | `oauth.grantType` | `"authorization_code"` (default) or `"client_credentials"` for non-interactive machine auth |
+| `oauth.callbackPort` | Callback port for pre-registered OAuth clients that require an exact redirect URI; overrides `MCP_OAUTH_CALLBACK_PORT` for this server |
 | `bearerToken` / `bearerTokenEnv` | Token or env var name; `bearerToken` supports `${VAR}` and `$env:VAR` interpolation |
 | `lifecycle` | `"lazy"` (default), `"eager"`, or `"keep-alive"` |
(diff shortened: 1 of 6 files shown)


webern (approved):
I need this... please merge.

nicobailon:
Implemented the callback-port use case on `main` in `fb53c85` through full `oauth.redirectUri` support rather than a separate numeric `callbackPort` option.

`oauth.redirectUri` covers the registered host, port, and path in one value, which avoids adding a second callback configuration surface. For example, Slack-style `http://localhost:3118/callback` can now be configured directly.

Closing as superseded by the broader redirect URI support. Thanks for the PR.

PR #97, opened 2026-05-13
fix: validate cached OAuth client redirect URI

mightymatth:
## Summary

When the local OAuth callback port changed between dynamic client registration and a later browser reauth attempt, some providers rejected the reused cached client with `Unregistered redirect_uri`; this PR stores registered redirect URIs and discards stale dynamic OAuth cache before starting that browser auth flow.

## Reproduction

1. Authenticate with an OAuth MCP server using dynamic client registration.
2. The adapter registers a dynamic OAuth client for the current callback URL, e.g. `http://localhost:19876/callback`.
3. Later, start OAuth auth again while that port is unavailable.
4. The callback server falls back to another port, e.g. `http://localhost:19877/callback`.
5. Before this PR, the adapter reused the cached dynamic `client_id` because cached client info was only scoped to `serverUrl`.
6. The authorization request then used the old `client_id` with the new `redirect_uri`.
7. OAuth providers that enforce registered redirect URIs reject that request with `invalid_request` / `Unregistered redirect_uri`.

## What changed

- Store the redirect URIs associated with dynamically registered OAuth clients.
- Before starting an interactive browser auth flow, compare the current callback URL with the cached dynamic client's redirect URIs.
- If the cached dynamic client was registered for a different or unknown callback URL:
  - clear cached client info
  - clear cached tokens
  - clear stale PKCE verifier
  - clear stale OAuth state
- Then continue with the normal SDK auth flow, which dynamically registers a fresh client for the current callback URL.

Refresh-token behavior is intentionally unchanged: cached client info is still returned normally from `clientInformation()`, because refresh grants do not use `redirect_uri`.

## Tests

- `npm test`
- `npm run test:oauth-provider`
- `npx tsc --noEmit`

### DIFF
--- a/__tests__/mcp-auth-flow-client-credentials.test.ts
+++ b/__tests__/mcp-auth-flow-client-credentials.test.ts
@@ -161,7 +161,10 @@ describe("mcp-auth-flow explicit auth", () => {
     const { authenticate } = await import("../mcp-auth-flow.ts");
     const { getOAuthState, updateClientInfo, updateTokens } = await import("../mcp-auth.ts");
 
-    updateClientInfo("expired", { clientId: "client" }, "https://api.example.com/mcp");
+    updateClientInfo("expired", {
+      clientId: "client",
+      redirectUris: ["http://localhost:19876/callback"],
+    }, "https://api.example.com/mcp");
     updateTokens("expired", {
       accessToken: "old-access",
       refreshToken: "old-refresh",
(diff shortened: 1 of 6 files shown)


nicobailon:
Implemented this on `main` in `fb53c85`.

Dynamic client registrations now store redirect URI metadata, and browser auth clears/re-registers cached dynamic clients when their stored redirect URIs are missing, malformed, or do not include the active callback URI. The refresh-token path intentionally still uses cached client info without applying this redirect check.

Closing as superseded by the mainline implementation. Thanks for the PR.

PR #100, opened 2026-05-14
fix: harden MCP server-name validation + ReDoS-safe regex search

av1155:
Two defensive fixes surfaced by a `semgrep` + `opengrep` + `madge` scan against v2.6.1, hardened after an adversarial review pass.

## 1. Path traversal in OAuth storage (`mcp-auth.ts`)

`getServerDir(serverName)` previously concatenated the server name into the auth-storage path via `join` without validation. MCP server names come from `mcp.json` so the threat model is low likelihood, but a stray `..` or absolute path would let an attacker write tokens outside `MCP_OAUTH_DIR`.

Two layers:

1. `assertSafeServerName` enforces an ASCII allow-list (`/^[A-Za-z0-9][A-Za-z0-9._:-]*$/`). The allow-list rejects path separators, lone `.`/`..`, NULs, control characters, whitespace, leading hyphens (CLI-flag confusion), and non-ASCII names (which would create distinct directories under NFC vs NFD normalisation on most Unix filesystems).
2. `getServerDir` resolves the candidate path then runs `path.relative` against the resolved base; anything that escapes (absolute path or starts with `..`) is rejected. The `path.relative` form tolerates the base being a filesystem root (POSIX `/`, Windows `C:\`), unlike a naive `startsWith(base + sep)` check.

Test coverage in `mcp-auth.test.ts` adds 17 traversal-shape inputs across three entry points (`saveAuthEntry`, `updateOAuthState`, `getAuthEntry`), plus 5 non-string inputs and 6 safe-name round-trip cases. 85 mcp-auth tests pass.

Rule: `javascript.lang.security.audit.path-traversal.path-join-resolve-traversal`.

## 2. ReDoS in regex-mode search (`proxy-modes.ts`)

`executeSearch(state, query, regex)` previously passed the caller's `query` straight into `new RegExp(query, "i")` when `regex=true`. A short alternation-overlap pattern like `(a|a)*b` (7 chars) can pin the regex engine for many seconds.

Three layers before compilation:

1. 256-char length cap stops trivially long inputs.
2. A throwaway `new RegExp` checks syntax. Without this, a malformed regex would be masked as `unsafe_pattern` by the ReDoS analyser instead of the accurate `invalid_pattern`.
3. `recheck` (the modern static ReDoS analyser) flags polynomial or exponential complexity, including the alternation-overlap patterns (`(a|a)*b`, `(a|ab)*c`, `^(a|aa)+$`, `(.|.)+a$`) that older heuristic checkers (for example `safe-regex`) miss.

The non-regex branch already metacharacter-escapes terms and is untouched.

Test coverage in `proxy-modes.test.ts` (new file): 8 cases covering the length cap, the `invalid_pattern` path for malformed regex, four catastrophic patterns rejected, a benign pattern accepted, and a non-regex-mode regression check.

Rule: `javascript.lang.security.audit.detect-non-literal-regexp.detect-non-literal-regexp`.

## Dependencies

Adds runtime dep `recheck@^4.5.0` (1 transitive: `synckit`). No native bindings, no build step. Synchronous `checkSync` API keeps `executeSearch` flat.

## Test wiring

The repo's `npm test` runs vitest, which only picks up `__tests__/**`. The root-level node:test files (existing and new) need their own runner. This PR adds two scripts without changing the existing `test` script:

```
test:node    runs all root-level node:test files (mcp-auth.test.ts,
             mcp-auth-flow.test.ts, mcp-callback-server.test.ts,
             mcp-oauth-provider.test.ts, proxy-modes.test.ts)
test:all     chains test (vitest) and test:node
```

CI can adopt `test:all` to cover both suites. I am open to moving the new tests under `__tests__/` with a vitest port instead if you prefer.

## Notes

Each fix is its own commit so the auth change and the search change can be reviewed independently.

The semgrep WARNING-level findings for those rules will remain after this PR lands because the rules match structurally on `path.resolve` and `new RegExp` with a non-literal argument; the static checker cannot follow the allow-list call or the `checkRegexSync` call. Two `nosemgrep` directives sit on the now-validated call sites so future scans flag only new regressions.

284 of 286 vitest tests still pass. The 2 failing ones (`interactive-visualizer-server.test.ts`) are pre-existing and depend on a missing `examples/interactive-visualizer/dist/server.js` build artifact, unrelated to this change.

Refs #101 (madge cycles, tracked separately).

### DIFF
--- a/mcp-auth.test.ts
+++ b/mcp-auth.test.ts
@@ -370,4 +370,91 @@ describe("mcp-auth", () => {
       assert.strictEqual(entry?.clientInfo?.clientId, "client")
     })
   })
+
+  describe("server name validation", () => {
+    const traversalNames: string[] = [
+      "",
+      ".",
+      "..",
+      "../etc",
+      "../../etc/passwd",
+      "/etc/passwd",
+      "\\windows\\system32",
+      "name/with/slash",
+      "name\\with\\backslash",
+      "name\0null",
+      // Whitespace and control characters: not exploitable on POSIX, but a
+      // typo footgun and unsafe on Windows / in log files.
+      " foo ",
+      "foo\nbar",
+      "foo\tbar",
+      "foo\rbar",
+      // Leading hyphen: looks like a CLI flag in any tool that re-uses the name.
+      "--help",
+      // Non-ASCII: NFC vs NFD would produce distinct directories on
+      // case-sensitive filesystems. Allow-list keeps us ASCII-only.
+      "café",
+      "café",
+    ]
+
+    for (const bad of traversalNames) {
+      // saveAuthEntry -> writeAuthEntry propagates throws; the validation surfaces here.
+      it(`should reject ${JSON.stringify(bad)} on save`, () => {
+        assert.throws(
+          () => saveAuthEntry(bad, { serverUrl: "https://example.com" }, "https://example.com"),
+          /Invalid MCP server name/,
+        )
+      })
+
+      // updateOAuthState propagates too; covers a second write path.
+      it(`should reject ${JSON.stringify(bad)} on updateOAuthState`, () => {
+        assert.throws(
+          () => updateOAuthState(bad, "state-value", "https://example.com"),
+          /Invalid MCP server name/,
+        )
+      })
+
+      // Read returns undefined (the try/catch in readAuthEntry swallows by design);
+      // crucially, no file outside TEST_DIR is created/read.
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks for the careful security pass here. Both important pieces are covered on main now: auth storage no longer uses raw server names as directory names, and regex search has a length cap plus the ReDoS safety check. Closing this PR since the current branch is stale against the newer implementation.

PR #107, opened 2026-05-18
feat: emit MCP lifecycle events on pi.events

sarukas:
## feat: emit MCP lifecycle events on `pi.events`

Adds a typed, versioned event contract on PI's shared event bus so
any extension / embedder can observe MCP server state without
adapter-specific glue or polling the `mcp` proxy tool.

### What

A new `mcp-events.ts` module emits at each existing MCP state
transition, under the reserved `mcp:` channel namespace, plus an
optional durable `mcp:state` mirror via `pi.appendEntry` for
consumers that attach mid-session:

| Channel | Payload | Phases / fields |
|---|---|---|
| `mcp:server` | `{ v, serverId, transport, at, phase, error? }` | `connecting`, `connected`, `disconnected`, `reconnecting`, `reconnected`, `idle_shutdown`, `errored` |
| `mcp:tools` | `{ v, serverId, transport, at, toolCount, toolNames?, source }` | `source`: `connect`\|`refresh`\|`cache`; `toolNames` omitted past 100 |
| `mcp:auth` | `{ v, serverId, transport, at, phase, authorizationUrl?, error? }` | `auth_required`, `auth_succeeded`, `auth_failed` |

Wired at the real transition seams already in the codebase:
`server-manager` (connect/close/error), `lifecycle`
(reconnect/idle — `close(name,{reason:"idle"})` distinguishes
`idle_shutdown` from `disconnected`), and `mcp-auth-flow` (OAuth).

### Properties

- **Additive & fire-and-forget.** No existing tool/command/UI
  behaviour changes. `dispatch()` is try/caught; a faulty
  subscriber/sink can never wedge a connection. `NOOP_EVENT_BUS` is
  the default until `init.ts` wires the real sink.
- **Versioned.** `v` is mandatory; additive fields within a version,
  `v` bumps on semantic change.
- **No secrets.** `authorizationUrl` is the only auth-adjacent field,
  only on the interactive `auth_required` phase.
- **Bounded.** `mcp:tools.toolNames` omitted past 100 tools.
- **Coalesced.** The durable `mcp:state` mirror skips the append when
  nothing observable changed (no per-no-op-reconnect JSONL growth).
- **Precise transport.** `lastTransportKind` is retained across
  disconnect so a `reconnecting` event for a previously-connected SSE
  server is not mislabelled `http`.
- **Testable.** `McpEventSink` is a structural type (tests use a
  plain spy, no PI runtime). Includes `__tests__/mcp-events.test.ts`,
  `__tests__/server-manager-events.test.ts`,
  `__tests__/state-snapshot.test.ts`.

### Why not hooks / why this shape

Hooks are agent-loop gates — wrong semantics for side-process status.
The event bus is the right channel for "an externally-managed
extension reports its own lifecycle." The contract is intentionally
extension-agnostic so embedders depend on the schema, not internals.

### Validated downstream

This contract was consumed end-to-end by a separate web embedder (a
PI-side observer subscribing on the shared loader `eventBus`, real
PI runtime, hermetic): `mcp:server`/`mcp:tools`/`mcp:auth` are
received and projected to a live status surface, replacing a fragile
proxy-tool poll. The schema needed no changes during that
integration.

### One open question for maintainers

`pi.appendEntry("mcp:state", …)` writes a `custom` session entry.
The README states this is "persisted state, not LLM context" — is an
unknown `customType` reliably excluded from the model context /
compaction across PI versions? If not, a churny multi-server config
could leak into the prompt. Happy to gate the durable mirror behind
an opt-in if that assumption isn't guaranteed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   envelope as a stable, additive integration contract. Each transition is also
   mirrored durably via `pi.appendEntry("mcp:state", …)` so consumers attaching
   mid-session can reconstruct state by replaying entries. Emission is
-  fire-and-forget and changes no existing tool/command/UI behaviour. See the
-  new "Events" section in the README.
+  fire-and-forget and changes no existing tool/command/UI behaviour. The
+  durable `mcp:state` write is coalesced (appended only when the
+  reconstructable state changes, not per transition) to avoid session-JSONL
+  write amplification, and the last observed transport kind is retained across
+  disconnect so `reconnecting` events are not mislabelled for SSE servers. See
+  the new "Events" section in the README.
 
 ## [2.6.1] - 2026-05-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Emit MCP lifecycle events on the shared `pi.events` bus under the reserved
+  `mcp:` namespace (`mcp:server`, `mcp:tools`, `mcp:auth`), with a versioned
+  envelope as a stable, additive integration contract. Each transition is also
+  mirrored durably via `pi.appendEntry("mcp:state", …)` so consumers attaching
+  mid-session can reconstruct state by replaying entries. Emission is
+  fire-and-forget and changes no existing tool/command/UI behaviour. See the
+  new "Events" section in the README.
+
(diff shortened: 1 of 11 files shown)


nicobailon:
Thanks for the contribution. I’m going to close this rather than ask for a rework here.

The idea is interesting, but the branch is stale against `main` and the implementation touches sensitive lifecycle/OAuth paths. In review, the durable `mcp:state` mirror can be written before the real manager/tool state is updated, and some auth-required cases can bypass the new event path. Since this would create a public event contract, I don’t want to merge a risky implementation as-is.

If we revisit MCP lifecycle events, I’d rather do it as a fresh, smaller change against current `main`.

PR #108, opened 2026-05-18
use 127.0.0.1 loopback IP literal instead of localhost for OAuth redirect URI

mcphailtom:
Per RFC 8252 §7.3, OAuth native-app clients should use the loopback IP literals (127.0.0.1, [::1]) for redirect URIs rather than the hostname 'localhost'.

This is a resubmission of #60, originally I encountered this problem with the datadog MCP but have just encountered the same issue with the official Slack MCP.


mcphailtom:
Closed for alternative approach which would cover this issue #114

PR #112, opened 2026-05-20
fix: prevent OAuth callback server port exhaustion

### LINKED ISSUE #104: OAuth callback server EADDRINUSE with multiple concurrent sessions
### Problem

When running many concurrent pi coding sessions, startup fails with:

```
Error: listen EADDRINUSE: address already in use ::1:19900
```

The full stack trace points to `ensureCallbackServer()` in `mcp-callback-server.ts:211`.

### Root Cause

Two issues compound:

1. **Eager port binding**: `initializeOAuth()` calls `ensureCallbackServer()` on every session startup, binding port 19876 (default) even when no OAuth flow will ever occur. With N concurrent sessions, N ports are consumed immediately.

2. **Sequential port scanning**: `ensureCallbackServer()` scans ports 19876-19900 sequentially when `strictPort` is false. With 25+ sessions, all ports in the range are exhausted and the function throws.

### Proposed Fix

**Two changes:**

**1. Make `initializeOAuth()` lazy** -- don't bind any port at startup. The callback server is already started on-demand inside `startAuth()` via `ensureCallbackServer()`, so the eager call in `initializeOAuth()` is unnecessary.

```typescript
// mcp-auth-flow.ts
export async function initializeOAuth(): Promise<void> {
  // No-op: callback server starts on demand via ensureCallbackServer()
  // inside startAuth(). This avoids EADDRINUSE when many sessions run
  // concurrently, each of which would otherwise bind a port at init.
}
```

**2. Use port 0 for non-strict binding** -- when `strictPort` is false (dynamic OAuth registration), use `listen(0)` to let the OS assign a guaranteed-free ephemeral port. This eliminates port scanning entirely and scales to any number of concurrent sessions. The redirect URI is already dynamically constructed from `getOAuthCallbackPort()`, so this is safe for dynamic registration flows.

```typescript
// mcp-callback-server.ts -- non-strict path
const candidateServer = createServer(handleRequest)
const assignedPort = await new Promise<number>((resolve, reject) => {
  candidateServer.once("error", reject)
  candidateServer.listen(0, "localhost", () => {
    const addr = candidateServer.address()
    resolve(typeof addr === "object" && addr ? addr.port : preferredPort)
  })
})
server = candidateServer
server.unref()
setOAuthCallbackPort(assignedPort)
```

When `strictPort` is true (pre-registered OAuth client with fixed redirect URI), the existing behavior is preserved: bind the exact configured port or throw.

### Impact

- Sessions that never use OAuth consume zero ports
- Sessions that do OAuth get an OS-assigned port instantly with no collision risk
- No limit on concurrent sessions
- Pre-registered OAuth clients (strictPort) unaffected

### Environment

- pi-mcp-adapter v2.6.1
- macOS, multiple concurrent pi sessions

cokekitten:
## Problem

Closes #104

Each Pi session eagerly binds an OAuth callback port (19876-19900) at startup via `initializeOAuth()` → `ensureCallbackServer()`. When processes are killed or exit abnormally, ports leak. After ~25 leaked ports, new Pi instances fail with:

```
Error: OAuth callback port 19876 is already in use and no free port was found in range 19876-19900
```

### Fix 1: Lazy initialization (`mcp-auth-flow.ts`)

`initializeOAuth()` is now a no-op. The callback server starts on demand inside `startAuth()` via `ensureCallbackServer()`. Sessions that never use OAuth consume zero ports.

### Fix 2: OS-assigned ephemeral port (`mcp-callback-server.ts`)

Non-strict mode now uses `listen(0)` to let the OS assign a guaranteed-free ephemeral port, replacing the sequential scan of ports 19876-19900. This:

- Eliminates the 25-session concurrency ceiling
- Removes the need for `MAX_PORT_SCAN_ATTEMPTS`
- Zero collision risk regardless of concurrent session count

`strictPort` mode (pre-registered OAuth clients that need an exact redirect URI) preserves the existing behavior: bind the configured port or throw.

### Fix 3: Process exit handlers (`mcp-callback-server.ts`)

Registers `SIGTERM`/`SIGINT`/`SIGHUP` signal handlers and a `process.on("exit")` fallback to synchronously close the HTTP server when the process is killed or exits abnormally. This prevents port leaks from stale listeners even with fix 1 + fix 2.

## Testing

All 284 tests pass (including the 4 updated callback-server tests):

```
32 test files, 284 tests passed
```

## Changes

| File | Change |
|------|--------|
| `mcp-auth-flow.ts` | `initializeOAuth()` → no-op (lazy) |
| `mcp-callback-server.ts` | Non-strict: `listen(0)` instead of sequential scan; register exit handlers |
| `__tests__/mcp-callback-server-unref.test.ts` | Updated mocks and assertions for new behavior |

### DIFF
--- a/__tests__/mcp-callback-server-unref.test.ts
+++ b/__tests__/mcp-callback-server-unref.test.ts
@@ -39,6 +39,7 @@ const mocks = vi.hoisted(() => {
       }),
       close: vi.fn((cb?: () => void) => cb?.()),
       unref: vi.fn(),
+      address: vi.fn(() => ({ port: 4338, family: "IPv6", address: "::1" })),
     };
 
     runtime.servers.push(server);
@@ -88,7 +89,8 @@ describe("mcp-callback-server", () => {
 
     await ensureCallbackServer();
 
-    expect(mocks.runtime.servers[0]?.listen).toHaveBeenCalledWith(4337, "localhost", expect.any(Function));
+    // Non-strict mode uses listen(0) for OS-assigned port
+    expect(mocks.runtime.servers[0]?.listen).toHaveBeenCalledWith(0, "localhost", expect.any(Function));
     expect(mocks.runtime.servers[0]?.unref).toHaveBeenCalledTimes(1);
   });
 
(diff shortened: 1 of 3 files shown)


rtfmkiesel:
Looking forward to this merge. This fixes an issue I discovered this week. 

If you run pi in a jail were binding to ports is not allowed, it would crash at startup even if none of my MCP servers use OAuth:

```
MCP OAuth initialization failed: Error: listen EACCES: permission denied 127.0.0.1:19876
    at Server.setupListenHandle [as _listen2] (node:net:1918:21)
    at listenInCluster (node:net:1997:12)
    at GetAddrInfoReqWrap.callback (node:net:2206:7)                                             
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:134:8) {
  code: 'EACCES',
  errno: -13,
  syscall: 'listen',
  address: '127.0.0.1',
  port: 19876
}
```

The hot fix I used was:
```sh
sed -i 's/await ensureCallbackServer()/\/\/ await ensureCallbackServer()/' ~/.pi/agent/npm/node_modules/pi-mcp-adapter/mcp-auth-flow.ts
```

nicobailon:
Thanks for digging into this one. The OAuth callback server has since been reworked on main so it starts lazily, uses an OS-assigned port when it can, preserves strict-port behavior for pre-registered clients, and unrefs/cleans up properly. That covers the port exhaustion issue, so I’m closing this stale PR.

PR #114, opened 2026-05-22
Add OAuth client metadata and redirect URI overrides

jonasherr:
## Summary

Adds optional per-server OAuth overrides for authorization-code flows:

- `oauth.redirectUri`
- `oauth.clientName`
- `oauth.clientUri`

When `oauth.redirectUri` is set, the OAuth provider uses that exact redirect URI for dynamic client registration and the callback server accepts that URI's path. Explicit redirect URIs use strict callback port binding, matching the existing behavior for pre-registered clients.

## Why

Some OAuth-protected MCP servers apply allowlisting or display/approval policy based on the registered OAuth client metadata or redirect URI. The current hardcoded values work for many servers, but some require a stable or client-compatible identity.

This keeps the current defaults unchanged while allowing users to configure compatible metadata when needed.

## Testing

- `npm run test:oauth-provider`
- `npx vitest run __tests__/mcp-callback-server-unref.test.ts __tests__/mcp-auth-flow-client-credentials.test.ts __tests__/config.test.ts`
- `npm test` mostly passed, but the existing `interactive-visualizer` dist artifact tests failed locally because `examples/interactive-visualizer/dist/*` is not present in this checkout.

### DIFF
--- a/OAUTH.md
+++ b/OAUTH.md
@@ -73,6 +73,9 @@ You can optionally provide a pre-registered client:
 - `oauth.clientId` - Pre-registered client ID (optional, SDK tries dynamic registration if not provided)
 - `oauth.clientSecret` - Client secret for confidential clients (optional)
 - `oauth.scope` - Requested OAuth scopes (optional)
+- `oauth.redirectUri` - Full redirect URI for authorization-code flow. Overrides the default `http://localhost:<port>/callback` callback URI.
+- `oauth.clientName` - Dynamic registration `client_name` override (default: `Pi Coding Agent`).
+- `oauth.clientUri` - Dynamic registration `client_uri` override (default: this package repository).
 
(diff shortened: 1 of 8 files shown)


mcphailtom:
I think this is probably a much better/broader solution to the problem I was fixing here #108

nicobailon:
Implemented this on `main` in `fb53c85`.

This includes the per-server `oauth.redirectUri`, `oauth.clientName`, and `oauth.clientUri` support from this PR, adapted on top of the lazy callback-server behavior from `#104`. It also binds explicit redirect URIs to the exact loopback host, port, and path, snapshots the redirect URI for the OAuth exchange, and adds coverage for the callback/metadata behavior.

Closing as superseded by the mainline implementation. Thanks for the PR.

PR #135, opened 2026-06-08
sanitize tool/resource descriptions to strip newlines

### LINKED ISSUE #134: Garbled display of mcp panel
When running on Windows, the MCP menu gets garbled.

Steps to reproduce:
- start /mcp 
- click on one of the installed mcp servers to expand its tools
- move cursor down and up to trigger list scrolling
- observe the output giving duplicate lines overwriting each other.

The end result is something like this:

Image

Expected result:
- up & down keys should scroll through the tools the same way it works under Linux and wsl

markokocic:
Fixes #134 

If description contains /r/n, the /r character would garble the output.
This adds sanitization of description

### DIFF
--- a/mcp-panel.ts
+++ b/mcp-panel.ts
@@ -73,6 +73,10 @@ function fuzzyScore(query: string, text: string): number {
   return qi === lq.length ? score : 0;
 }
 
+function sanitizeDesc(s: string | undefined | null): string {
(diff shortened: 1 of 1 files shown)


kamilmodest:
Thank you! I faced similar issue with [gitnexus](https://github.com/abhigyanpatwari/GitNexus)'s official MCP.

PR #138, opened 2026-06-10
fix: honor user keybindings for navigation in MCP panels

### LINKED ISSUE #137: MCP panels ignore custom keybindings (`tui.select.*` from `keybindings.json`)
## Problem

The interactive panels — the `/mcp` server/tool picker (`mcp-panel.ts`) and the `/mcp` setup flow (`mcp-setup-panel.ts`) — hardcode their navigation keys via `matchesKey(data, "up")`, `matchesKey(data, "down")`, `matchesKey(data, "return")`, etc. from `@earendil-works/pi-tui`.

They never consult the user's keybinding configuration (`~/.pi/agent/keybindings.json`), even though pi injects a `KeybindingsManager` into every `ctx.ui.custom()` factory precisely for this purpose. The adapter's factories currently receive it and discard it:

```ts
ctx.ui.custom(
  (tui, _theme, _keybindings, done) => {   // ← manager available but unused
    return createMcpPanel(config, cache, provenanceMap, callbacks, tui, ...);
  },
  ...
);
```

Per pi's extension docs (`extensions.md`):

> Custom editors and `ctx.ui.custom()` components receive `keybindings: KeybindingsManager` as an injected argument. They should use that injected manager directly…

## Reproduction

1. Configure emacs-style list navigation in `~/.pi/agent/keybindings.json`:

   ```json
   {
     "tui.select.up": ["up", "ctrl+p"],
     "tui.select.down": ["down", "ctrl+n"]
   }
   ```

2. `/reload` (or restart) pi. Verify ctrl+p/ctrl+n move the selection in pi core selectors (e.g. the model selector, `/resume` picker).
3. Open `/mcp` (or `/mcp-auth`, or the `/mcp setup` flow).
4. Press ctrl+n / ctrl+p.

**Expected:** selection moves down/up, same as in pi core selection lists.

**Actual:** nothing happens (worse, in the main `/mcp` panel a stray control character can end up being ignored only by luck of the printable-character check). Only the hardcoded arrow keys work.

## Environment

- pi-mcp-adapter 2.8.0 (also present at v2.9.0 / current `main` — the input handlers are unchanged)
- @earendil-works/pi-coding-agent 0.74.x, macOS, Terminal.app

## Proposed fix

Thread the injected `KeybindingsManager` from the three `ctx.ui.custom()` call sites in `commands.ts` into both panels, and resolve list-navigation keys through it:

- `matchesKey(data, "up")` → `keybindings.matches(data, "tui.select.up")`
- `matchesKey(data, "down")` → `keybindings.matches(data, "tui.select.down")`
- `matchesKey(data, "return")` (list confirm) → `keybindings.matches(data, "tui.select.confirm")` (defaults to `enter`, which `matchesKey` treats as an alias of `return`, so default behavior is unchanged)

with a fallback to the current hardcoded keys when no manager is supplied, so the panels remain constructible standalone (as the existing unit tests do).

`escape`/`ctrl+c` should probably stay hardcoded for now: the panels assign them distinct semantics (clear-search/go-back vs hard-quit), which doesn't map cleanly onto the single `tui.select.cancel` binding (whose default is `["escape", "ctrl+c"]`).

Happy to send a PR for this.

owenniles:
Fixes #137.

The `/mcp` panel and setup flow hardcoded `matchesKey(data, "up"/"down"/"return")`
and discarded the `KeybindingsManager` that pi injects into `ctx.ui.custom()`
factories, so custom `tui.select.*` bindings (e.g. emacs-style ctrl+p/ctrl+n)
worked in pi core selectors but not here.

- `panel-keys.ts`: `createPanelKeys(keybindings?)` resolves up/down/confirm
  through the user's `tui.select.*` bindings, falling back to the previous
  hardcoded keys when no manager is supplied (panels stay constructible
  standalone, as existing tests do). The manager is typed as a minimal
  structural interface so the panels don't depend on the concrete class.
- `commands.ts`: all three `ctx.ui.custom()` call sites now pass the
  previously discarded `keybindings` argument into the panel options.
- `escape`/`ctrl+c` stay hardcoded: the panels give them distinct semantics
  (back/clear-search vs hard-quit) that don't decompose onto the single
  `tui.select.cancel` binding.

No behavior change for default keybindings: `tui.select.confirm` defaults to
`enter` (alias of `return`), up/down default to the arrow keys.

Testing: `npx tsc --noEmit` clean; new `__tests__/mcp-panel-keybindings.test.ts`
(7 tests, real `KeybindingsManager`) covers custom bindings, defaults,
rebinding that removes a default, and the no-manager fallback. Full suite:
327 passed, 2 failed — both `interactive-visualizer-server.test.ts`, failing
identically on `main` (needs a built example; unrelated). Manually verified
ctrl+p/ctrl+n in `/mcp`, `/mcp-auth`, and the setup flow.

### DIFF
--- a/__tests__/mcp-panel-keybindings.test.ts
+++ b/__tests__/mcp-panel-keybindings.test.ts
@@ -1,0 +1,186 @@
+import { KeybindingsManager, TUI_KEYBINDINGS } from "@earendil-works/pi-tui";
+import { describe, expect, it, vi } from "vitest";
+import { createMcpPanel } from "../mcp-panel.ts";
+import { createMcpSetupPanel, type SetupPanelCallbacks } from "../mcp-setup-panel.ts";
+import { createPanelKeys } from "../panel-keys.ts";
+import type { McpDiscoverySummary } from "../config.ts";
+import type { McpConfig, McpPanelCallbacks } from "../types.ts";
+
+const CTRL_P = "\x10";
+const CTRL_N = "\x0e";
+const UP = "\x1b[A";
+const DOWN = "\x1b[B";
+const ENTER = "\r";
+
+function createEmacsKeybindings(): KeybindingsManager {
+  return new KeybindingsManager(TUI_KEYBINDINGS, {
+    "tui.select.up": ["up", "ctrl+p"],
+    "tui.select.down": ["down", "ctrl+n"],
+  });
+}
+
+function createTwoServerConfig(): McpConfig {
+  return {
+    mcpServers: {
+      alpha: { url: "https://alpha.example.com/mcp", auth: "oauth" },
+      beta: { url: "https://beta.example.com/mcp", auth: "oauth" },
+    },
+  };
+}
+
+function createAuthCallbacks(): McpPanelCallbacks {
+  return {
+    reconnect: async () => true,
+    canAuthenticate: () => true,
+    authenticate: vi.fn(async () => ({ ok: true })),
+    getConnectionStatus: () => "needs-auth",
+    refreshCacheAfterReconnect: () => null,
+  };
+}
+
+function createEmptyDiscovery(): McpDiscoverySummary {
+  return {
+    sources: [],
+    imports: [],
+    hasAnyConfig: false,
+    hasAnyDetectedPaths: false,
+    hasSharedServers: false,
+    hasPiOwnedServers: false,
+    totalServerCount: 0,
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks @owenniles, this is merged. I also added an Unreleased changelog note crediting PR #138.

PR #140, opened 2026-06-11
fix: oauth scopes and inject auth params

### LINKED ISSUE #133: OAuth: scope parameter not included in authorization URL
## Problem

When using OAuth authentication with MCP servers that require specific scopes, the OAuth authorization URL is generated **without the `scope` parameter**, even when scope is provided in the configuration. This causes the server to reject subsequent MCP requests with `insufficient_scope` error.

### Error Example

```
{"error":"insufficient_scope","required":["some-scope"]}
```

Even though the scope was specified in the OAuth config, it wasn't included in the authorization request sent to the OAuth server.

## Root Cause

The `clientMetadata` getter in `mcp-oauth-provider.ts` didn't include the `scope` property, so the MCP SDK couldn't properly handle scope resolution during authorization.

## Solution

Add `scope` to clientMetadata

In `mcp-oauth-provider.ts`, add `scope` to the client metadata returned for authorization_code flow:

```typescript
// mcp-oauth-provider.ts line ~135
return {
  redirect_uris: [redirectUrl],
  client_name: this.config.clientName ?? "Pi Coding Agent",
  client_uri: this.config.clientUri ?? "https://github.com/nicobailon/pi-mcp-adapter",
  grant_types: ["authorization_code", "refresh_token"],
  response_types: ["code"],
  token_endpoint_auth_method: this.config.clientSecret ? "client_secret_post" : "none",
  scope: this.config.scope,  // <-- ADD THIS LINE
}
```

This ensures the MCP SDK properly includes the scope in the authorization URL.

carlosdagos:
should fix #133 (issue I also had)

by using the hook `addClientAuthentication` we can supply the scopes as set out in user configs, otherwise they can get an error by the idp saying scopes are empty

also if the client_id is available, inject it too, as it seems to be required by idps such as entra

### DIFF
--- a/__tests__/mcp-oauth-provider.test.ts
+++ b/__tests__/mcp-oauth-provider.test.ts
@@ -6,6 +6,92 @@ import { UnauthorizedError } from "@modelcontextprotocol/sdk/client/auth.js";
 import { McpOAuthProvider } from "../mcp-oauth-provider.ts";
 import { saveAuthEntry, updateOAuthState } from "../mcp-auth.ts";
 
+describe("McpOAuthProvider clientMetadata scope", () => {
+  it("includes scope in authorization_code clientMetadata when configured", () => {
+    const provider = new McpOAuthProvider(
+      "scope-test",
+      "https://api.example.com/mcp",
+      { scope: "api://resource/.default openid" },
+      { onRedirect: async () => {} },
+    )
+    expect(provider.clientMetadata.scope).toBe("api://resource/.default openid")
+  })
+
+  it("omits scope from clientMetadata when not configured", () => {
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks @carlosdagos. I picked up the scope fix from this PR and landed it in #142 / 9603544 with one adjustment: the token request hook now preserves the SDK token endpoint auth-method selection, so `client_secret_basic` providers still get Basic auth while configured scopes are sent for authorization-code flows.

I also credited you in the changelog. Closing this PR since the fix is now on main.

carlosdagos:
Thank you mate!

PR #144, opened 2026-06-17
fix: strip additionalProperties from direct MCP tool schemas

comtihon:
## Summary

Strips `additionalProperties: false` from MCP tool input schemas when registering direct tools with pi.

MCP servers can declare `additionalProperties: false` in their tool schemas. Pi validates tool call arguments against these schemas **before** the `tool_call` extension hook fires, so any plugin that needs to inject extra fields into tool call args (e.g. `post_compact`, tracing metadata) triggers a validation failure on strict-schema tools — the tool never runs and the hook never fires.

## Implementation

- **`utils.ts`** — add exported `stripAdditionalProperties(schema)` helper that omits `additionalProperties` from the schema object.
- **`index.ts`** — call `stripAdditionalProperties` on `spec.inputSchema` before wrapping with `Type.Unsafe` when registering each direct tool. Import the helper from `utils.ts`.

## Tests

- New `__tests__/strip-additional-properties.test.ts` covering: normal schema passthrough (no `additionalProperties`) and stripping `additionalProperties: false` from a strict schema.
- Full suite passes (`tsc --noEmit` clean; pre-existing `interactive-visualizer-server` failures are unrelated — they expect a built example artifact not present in the checkout).

## Usage

No config change needed — the stripping is unconditional for all direct tool schemas.

### DIFF
--- a/__tests__/strip-additional-properties.test.ts
+++ b/__tests__/strip-additional-properties.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it } from "vitest";
+import { stripAdditionalProperties } from "../utils.ts";
+
+describe("stripAdditionalProperties", () => {
+  it("passes through a normal schema unchanged", () => {
+    const schema = {
+      type: "object",
+      properties: {
+        name: { type: "string" },
+        count: { type: "number" },
+      },
+      required: ["name"],
+    };
+
+    const result = stripAdditionalProperties(schema);
+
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks @comtihon. This was handled through #162 and shipped in v2.11.0, and I kept your PR #144 credit in the changelog. Closing this as superseded.

Issue #18, opened 2026-03-16
Several improvement suggestions: tool rendering, timeout param, abort signal


igor-makarov:
Hi,

First of all, big fan of your many many extensions.

I've made some changes in [my fork](https://github.com/igor-makarov/pi-mcp-adapter) of the MCP extension. 

I didn't want to submit unsolicited PRs so I'm making a list of changes I've done. Hoping you read this.

Changes is my fork:

* Propagate abort signal to the MCP client, so that `Esc` key aborts the call
* Render tool title with its params
* Render tool output in the style of `bash` tool, with truncation and `Ctrl-O` expansion
* Support `timeout` param and pass it to the MCP client
* Truncate output and write full output to a file if exceeding 50KB

Let me know if you want these upstreamed, and if so, how.

Thanks!

Edit: Added one more row about output truncation.

joeldrotleff:
Hey @igor-makarov just want other say I was having the issue with full output showing up and your fork is working great for me, kudos

bjufre:
@igor-makarov @joeldrotleff I would really appreciate this, because right now some `mcp` outputs are massive and causes the session to hang and then I cannot reopen the session 🙏

nicobailon:
Closing this as completed across the later releases: abort propagation, richer call rendering, request timeouts, and output guarding/truncation are all now implemented.

PR #149, opened 2026-06-22
Forward agent abort signal so Esc/Ctrl+C cancels MCP tool calls (fixes #40)

### LINKED ISSUE #40: Hitting escape doesn't cancel execution of an mcp

murrayju:
## Summary

Pressing <kbd>Esc</kbd>/<kbd>Ctrl+C</kbd> in Pi doesn't cancel an in-flight MCP tool call — the call runs to completion and Pi discards the (late) result. This is especially painful for long-running tools, which block the agent until they finish or time out.

Fixes #40 (also addresses the "propagate abort signal" item in #18).

## Root cause

Pi passes a per-turn `AbortSignal` into every tool's `execute(toolCallId, params, signal, ...)` and aborts it on <kbd>Esc</kbd>/<kbd>Ctrl+C</kbd>. The MCP SDK already knows how to act on this: when `callTool`/`readResource` are given `RequestOptions.signal`, an abort triggers a `notifications/cancelled` JSON-RPC notification to the server.

But the adapter dropped the signal on the floor in all three execution paths:

- **`index.ts`** — the gateway `mcp` tool's `execute` ignored its signal (`}, _signal, _onUpdate, _ctx) {`) and didn't pass it to `executeCall`.
- **`proxy-modes.ts`** — `executeCall` had no signal parameter, and its `callTool`/`readResource` calls passed no `RequestOptions`.
- **`direct-tools.ts`** — the executor wrapper `return async function execute(_toolCallId, params)` didn't even declare the signal arg, and its `callTool`/`readResource` calls passed no `RequestOptions`.

Because no `signal` reached the SDK, `notifications/cancelled` was never sent, so the server kept working and Pi ignored the eventual result.

## Changes

- Capture the abort `signal` in the gateway `mcp` tool (`index.ts`) and the direct-tool executor (`direct-tools.ts`).
- Add a `signal?: AbortSignal` parameter to `executeCall` (`proxy-modes.ts`) and thread it through from `index.ts`.
- Forward the signal as `RequestOptions` to `callTool` (`{...}, undefined, { signal }`) and `readResource` (`{ uri }, { signal }`) in both the proxy and direct-tool paths.

The browser-driven `ui-server.ts` `callTool` is intentionally left alone — it has no Pi `AbortSignal` in scope and already has its own browser "Cancel" flow.

## Testing

- Added `__tests__/abort-signal-forwarding.test.ts` covering all three: `executeCall` → `callTool`, `executeCall` → `readResource` (resource tools), and `createDirectToolExecutor` → `callTool`. Each asserts the exact `RequestOptions.signal` is forwarded.
- `npx tsc --noEmit` passes.
- Full suite: 373 pre-existing tests pass + 3 new. The only failures (`interactive-visualizer-server.test.ts`, 2 tests) are pre-existing and unrelated — they read a build artifact (`examples/interactive-visualizer/dist/`) that isn't generated without a build step, and fail identically on a clean `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Forwarded the agent's abort signal to MCP `callTool`/`readResource` requests so pressing `Esc`/`Ctrl+C` cancels in-flight tool calls (the SDK now emits `notifications/cancelled`) instead of running them to completion. Fixes #40.
+
 ## [2.10.0] - 2026-06-13
 
 ### Added
--- a/__tests__/abort-signal-forwarding.test.ts
+++ b/__tests__/abort-signal-forwarding.test.ts
@@ -1,0 +1,158 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// These tests lock in that the agent's AbortSignal (raised when the user hits
+// Esc / Ctrl+C) is forwarded to the MCP SDK's callTool/readResource as
+// RequestOptions.signal. Without it, the SDK never emits notifications/cancelled
+// and long-running tool calls keep running after the user cancels (issue #40).
+
+const mocks = vi.hoisted(() => ({
+  lazyConnect: vi.fn(),
+  getFailureAgeSeconds: vi.fn(),
+}));
+
+vi.mock("../init.ts", () => ({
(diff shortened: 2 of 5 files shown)


nicobailon:
Thanks for the abort-signal work, Justin. PR #159 and current `main` now carry this through connect, tool, and resource calls with broader cancellation coverage, and issue #40 is complete, so this patch is superseded. I added your #149 credit to the changelog before closing this.

PR #152, opened 2026-06-24
feat: default stdio server cwd to the session cwd

TimoFreiberg:
Problem: Stdio MCP servers without an explicit `cwd` inherited the host process's `process.cwd()`. When pi is embedded in another app that runs many sessions in one process (e.g. a GUI host), process.cwd() is the host's dir, not the agent session's. Server-written files (e.g. playwright screenshots, which save relative to the server cwd) then land in the host dir instead of the session's working directory, where the agent can't find them.

Fix: Pass the session cwd (`ctx.cwd`) into `McpServerManager` and use it as the default spawn cwd when a server config doesn't set one. Explicit `cwd` in config still wins.

I need this for my WIP pi gui


nicobailon:
Thanks Timo. I merged this and kept your changelog credit for PR #152. The session-cwd default is in main now.

Issue #121, opened 2026-05-27
OAuth flow auto-opens browser without user consent on every MCP reconnect


nekrut:
## Description

When an MCP server requires OAuth (e.g. a remote server returning 401), `pi-mcp-adapter` v2.4.0 immediately calls `open(authorizationUrl)` in `mcp-auth-flow.ts` without any user interaction or consent prompt. This opens the system browser automatically.

This happens on **every reconnect** — each time the brain restarts or starts a new session, the adapter re-initializes the MCP connection, gets a 401 from the server, and silently opens the browser again. The result is the browser being opened repeatedly (observed every turn with the `brc-analytics` MCP server at `https://dev.brc-analytics.org/api/v1/mcp/`).

## Reproduction

1. Configure an MCP server that requires OAuth (e.g. `brc-analytics`) in `mcp.json`
2. Start the agent (Loom/Orbit)
3. Browser opens automatically with `http://localhost:<random-port>/?session=<uuid>` on every session start

## Expected behavior

The OAuth flow should only be triggered by explicit user action (e.g. a button click or `/connect` command), not silently on every reconnect.

## Observed behavior

`xdg-open` is called by `node loom.js --mode rpc` on every session start. The browser opens repeatedly and the OAuth callback page (`http://localhost:<port>/?session=<uuid>`) continuously refreshes without completing the flow.

## Relevant code

`mcp-auth-flow.ts` line ~214:
```ts
await open(authorizationUrl)
```

This is called unconditionally inside `performAuth()` with no user gate.

## Workaround

Set `"disabled": true` on the affected MCP server entry in `mcp.json`.

## Environment

- pi-mcp-adapter: 2.4.0
- Platform: Linux (Ubuntu 24.04)
- MCP server: `https://dev.brc-analytics.org/api/v1/mcp/`

nicobailon:
Closing this as fixed by the current OAuth flow. Browser auth is explicit or gated by `settings.autoAuth === true`; reconnects no longer auto-open the browser by default.

PR #172, opened 2026-07-06
feat: MCP_UI_VIEWER=none to suppress UI window while keeping tool result

stevekrouse:
## Summary

Adds a new `MCP_UI_VIEWER=none` value (also accepts `off` / `disabled`) that runs MCP UI tools and their backing UI server while **suppressing the browser/Glimpse window**.

## Why

Currently `MCP_UI_VIEWER` only accepts `browser` or `glimpse` — there's no way to call a UI tool and get its inline `content` result back to the agent without a window popping open. That's a problem for:

- **Headless / CI / remote sessions** where there's no display or you don't want focus-stealing windows
- **Users who prefer inline output** — they want the tool's text result delivered to the conversation, not a browser tab
- **Agent-driven workflows** where the agent calls the tool for its result but the human doesn't need (or want) the interactive UI

Related sentiment is in #121 (auto-open browser without consent), but that issue is specifically about the *OAuth* auth flow (`mcp-auth-flow.ts`). This PR addresses the analogous problem for the MCP **UI tool** path in `ui-session.ts`.

## How it works

When `MCP_UI_VIEWER` is `none` / `off` / `disabled`, `maybeStartUiSession()` now:
- ✅ Still starts the internal UI server (so streaming, `structuredContent`, and the tool's inline result keep working)
- ✅ Still returns the `UiSessionRuntime` so the tool call proceeds normally
- ❌ Does **not** call `openInBrowser()` or `openGlimpseWindow()` — no window opens
- 🔔 Emits a one-line `info` notification with the UI URL, so it can still be opened manually if ever needed

The viewer selection logic is refactored so the new suppressed branch short-circuits before the Glimpse/browser branch. Behavior for `browser` / `glimpse` / unset is unchanged.

## Usage

```bash
# Suppress the window; tool still runs and returns inline result
export MCP_UI_VIEWER=none
pi
```

## Checklist

- [x] Code change in `ui-session.ts`
- [x] README docs updated (the "Native rendering" paragraph)
- [x] Existing UI test suite passes (`ui-integration`, `ui-session-messages`, `ui-streaming` — 36/36)
- [x] No new TypeScript errors introduced (verified against `tsc --noEmit`)

## Notes / Alternatives considered

- I deliberately did **not** add a `settings.ui: false` knob in `McpSettings` — the existing viewer selection is already env-var-driven (`MCP_UI_VIEWER`), so extending that is the smallest, most consistent change. Happy to also add a `settings.ui`/`settings.uiViewer` config key if you'd prefer the JSON-config route.
- I considered also suppressing the "📺 Interactive UI is now open in your browser" suffix message that `proxy-modes.ts` appends to the tool result when `uiResourceUri` is set. I left it as-is for now to keep this change minimal and focused on the window-open behavior, but I can update that message (or skip it when suppressed) in this PR if you'd like.

## Testing

```
__tests__/ui-integration.test.ts  ✓ 10
__tests__/ui-session-messages.test.ts ✓ 6
__tests__/ui-streaming.test.ts ✓ 20
Test Files  3 passed (3)
     Tests  36 passed (36)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `MCP_UI_VIEWER=none` / `off` / `disabled` to suppress MCP UI browser or Glimpse windows while keeping inline tool results available. Thanks @stevekrouse for PR #172.
 - Surfaced MCP server `instructions` from the initialize handshake: captured at connect time, cached alongside tool metadata, shown as a truncated head in the `mcp` proxy tool description, previewed in `mcp({ server: "name" })` listings, and available in full via the new `mcp({ instructions: "name" })` mode. Thanks @JeongJuhyeon for issue #188 and PR #189.
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ MCP servers can ship interactive UIs via the [MCP UI](https://github.com/MCP-UI-
 3. pi-mcp-adapter fetches the UI HTML and opens it in an iframe
 4. The UI can call MCP tools and send messages back to the agent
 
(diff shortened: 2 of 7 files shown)


stevekrouse:
Thanks for fixing & merging my slop PR 😅

Issue #173, opened 2026-07-06
MCP server resource metadata URL is not used during authorization flow


jameswarren:
**Description:** As part of [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728#section-5.1), the response from a MCP server to an unauthenticated request may use the `www-authenticate` header to provide a resource metadata URL to the client. The current implementation does not forward the metadata URL to the MCP SDK's auth flow, preventing a successful connection if the metadata contains a separate authorization server.

**Scenario:** I encountered this behavior while attempting to connect to an MCP server hosted on an AWS Bedrock AgentCore Runtime (which requires an external authorization server for OAuth). 

* The URL for an MCP server hosted in AgentCore is of the form `https://bedrock-agentcore.<REGION>.amazonaws.com/runtimes/arn%3Aaws%3Abedrock-agentcore%3A<REGION>%3A<ACCOUNT_ID>%3Aruntime%2F<SERVER>/invocations?qualifier=DEFAULT'`

* An unauthenticated `initialize` request returns a `401` response with the `www-authenticate` header according to RFC 9728:

```
$ curl -si -X POST \
  'https://bedrock-agentcore.<REGION>.amazonaws.com/runtimes/arn%3Aaws%3Abedrock-agentcore%3Aus-west-2%3A<ACCOUNT_ID>%3Aruntime%2F<SERVER>/invocations?qualifier=DEFAULT' \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}' 2>&1

HTTP/2 401 
date: Tue, 09 Jun 2026 01:03:17 GMT
content-type: application/json
x-amzn-requestid: 409d3917-121c-490a-8d1b-d1342a6e8441
x-amzn-errortype: MissingAuthenticationTokenException:http://internal.amazon.com/coral/com.amazon.coral.service/
www-authenticate: Bearer resource_metadata="https://bedrock-agentcore.<REGION>.amazonaws.com/runtimes/arn%3Aaws%3Abedrock-agentcore%3Aus-west-2%3A<ACCOUNT_ID>%3Aruntime%2F<SERVER>/invocations/.well-known/oauth-protected-resource?qualifier=DEFAULT"

{"jsonrpc":"2.0","error":{"code":-32001,"message":"Missing Authentication Token"}}
```

* The resource metadata URL contains the authorization servers for the MCP server:

```
[...]

PR #174, opened 2026-07-06
fix: pass resource_metadata URL to MCP SDK auth

jameswarren:
Details in #173. Uses an independent probe to determine if the resource uses the `www-authenticate` header to return a resource metadata URL. If present, forwards the URL to the underlying MCP SDK auth functionality.


nicobailon:
Thanks James. I replaced this on main in 60d9003 with the same idea: use the server-advertised protected-resource metadata during OAuth, preserve the advertised scope, and complete auth through the SDK directly. You’re credited in CHANGELOG for issue #173 and PR #174, so I’m closing this PR as superseded.

PR #177, opened 2026-07-13
Fix collapsing single-line MCP output

maxpaulus43:
## Problem
- My mcp servers return payloads in huge one-line minified json strings. The mcp adapter extension doesn't truncate that because it takes up only a single line.

## Summary
- collapse single-line MCP result text (such as minified JSON)
- keep the full result available with Ctrl+O
- cover the behavior with a regression test

## Testing
- `git diff --check`
- Not run: `npm test -- --run __tests__/tool-result-renderer.test.ts` (`vitest` is not installed in this checkout)
- manually ran some mcp commands that return minified json with no newlines, and the result is properly truncated


nicobailon:
Thanks for tackling the long single-line output issue, Max. The width-aware renderer from #181 and #209 now covers this on current `main` and keeps the full output available when expanded, so this patch is superseded. I added your #177 credit to the changelog before closing this.

PR #182, opened 2026-07-15
fix(config): bind server auth to url in mergeServerMaps (credential exfiltration)

ductiletoaster:
## Summary

`mergeServerMaps` (config.ts) merges same-named MCP server entries **per-field** across config-precedence sources:

```js
merged[name] = { ...(merged[name] ?? {}), ...definition };
```

A higher-precedence config source that supplies **only a new `url`** for an already-defined server therefore **retains the lower-precedence entry's auth material** and applies it to the new url:

```
base (lower):     { url: A, headers: { Authorization: "Bearer ${TOKEN}" } }
override (higher):{ url: B }
merged:           { url: B, headers: { Authorization: "Bearer ${TOKEN}" } }   // token now sent to B
```

`resolveHeaders` → `interpolateEnvRecord` then expands `${TOKEN}` from `process.env` and ships the real credential as `Authorization` to whatever the merged `url` ends up being (server-manager.ts).

## Why this is a security issue

Config precedence is `~/.config/mcp/mcp.json` (shared-global, lowest) → `~/.pi/agent/mcp.json` (pi-global) → `<cwd>/.mcp.json` (project) → `<cwd>/.pi/mcp.json` (project-pi, highest). When a **less-trusted** higher-precedence source (a workspace `.mcp.json` from a cloned repo, or `~/.pi/agent/mcp.json` written by a prompt-injected agent) redefines only the `url` of a server whose credentials were baked into a **more-trusted** lower-precedence source, the per-field merge exfiltrates those credentials to an attacker-controlled endpoint. The credential is bound to *a name*, not to *the url it was issued for*.

## The fix

Bind credentials to the url that supplied them. In `mergeServerMaps`, when a higher-precedence override changes an existing server's `url`, drop the inherited url-bound auth fields — `headers`, `bearerToken`, `bearerTokenEnv`, `oauth` — before merging. Auth that the override itself re-supplies still applies (it is spread last).

Behaviour is **unchanged** when:
- the url is identical across sources (partial overrides still inherit auth), or
- the override omits `url` entirely (partial overrides still inherit).

Only the "higher-precedence source repoints the server at a different url" case changes — and there, inherited credentials are dropped rather than forwarded.

## Test plan

Added to `__tests__/config.test.ts` (all green; full suite unaffected apart from the pre-existing `interactive-visualizer` example-artifact tests):

- [x] same url + partial override → base headers preserved
- [x] url changed, no new auth → no `Authorization` on the merged result
- [x] url changed + new headers → only the new headers survive
- [x] exfiltration scenario: baked `{url:A, headers:{Authorization:${TOKEN}}}` + url-only override `{url:B}` → merged carries no token
- [x] `tsc --noEmit` clean

### DIFF
--- a/__tests__/config.test.ts
+++ b/__tests__/config.test.ts
@@ -264,6 +264,81 @@ describe("config discovery", () => {
     expect(JSON.stringify(entry)).not.toContain("LITELLM_API_KEY");
   });
 
+  // Per-field coverage for every member of URL_BOUND_AUTH_FIELDS beyond
+  // headers/bearerTokenEnv — guards against a regression that silently drops a
+  // field from the strip set while the other tests stay green.
+  it("drops an inherited bearerToken when a url-only override repoints the server", async () => {
+    const home = mkdtempSync(join(tmpdir(), "pi-mcp-urlauth-bt-home-"));
+    const project = mkdtempSync(join(tmpdir(), "pi-mcp-urlauth-bt-project-"));
+    writeBakedAndOverride(
+      home,
+      project,
+      { url: URL_A, bearerToken: "secret-bearer-token" },
+      { url: URL_B },
+    );
+
+    const { loadMcpConfig } = await import("../config.ts");
+    const config = loadMcpConfig();
+
+    const entry = config.mcpServers.litellm;
+    expect(entry).toEqual({ url: URL_B });
+    expect(entry.bearerToken).toBeUndefined();
+    expect(JSON.stringify(entry)).not.toContain("secret-bearer-token");
+  });
+
+  it("drops inherited oauth config when a url-only override repoints the server", async () => {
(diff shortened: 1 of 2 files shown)


ductiletoaster:
Expanded test coverage (pushed to this branch):

- Per-field stripping for the remaining credential fields — a url-only override now has explicit tests that it drops an inherited `bearerToken` and an inherited `oauth` config (incl. `clientSecret`), not just `headers`/`bearerTokenEnv`. Guards against a regression that silently removes a field from the strip set.
- A three-source laundering test: `{url:A, headers:VK}` → `{headers:VK}` (no url) → `{url:B}` (no auth), folded via `loadMcpConfig`. Confirms the strip decision keys off the accumulated (folded) entry's url, not just a pairwise base.

Full suite green; `tsc --noEmit` clean.

PR #186, opened 2026-07-16
fix: make footer status actionable, not an inventory ratio

### LINKED ISSUE #93: updateStatusBar shows 0/N for lazy servers before first tool call
## Describe the issue
The MCP status bar shows "MCP: 0/N servers" for lazy lifecycle servers until a tool call triggers the connection, then updates to the correct count. After idle shutdown it reverts to 0 again.

`updateStatusBar` in `init.ts` line ~278 uses `getAllConnections().size` to count connected servers. For lazy servers, no connection entry exists in the Map until `connect()` is called, so the count is 0. After idle shutdown, `manager.close()` deletes the entry entirely, so it goes back to 0.

The `/mcp` status command in `proxy-modes.ts` handles this differently:
```
const connectedCount = servers.filter(s => s.status === "connected").length;
```

## To Reproduce
1. Configure a server with `"lifecycle": "lazy"` in mcp.json
2. Start pi, status bar shows "MCP: 0/1 servers"
3. Call any tool from that server, status updates to "MCP: 1/1 servers"
4. Wait for idle timeout, status reverts to "MCP: 0/1 servers"

Not sure if this is intentional. The count is technically accurate since no active connection exists, but it reads like the server is broken rather than just not yet connected.

Boubalou:
## Summary

Fixes #93.

With the default `lazy` lifecycle, servers start disconnected by design. The footer permanently showed `MCP: 0/N servers` in accent color, which reads as broken/incomplete rather than healthy idle.

**New behavior** (show the exception, not the inventory):

| State | Footer |
|-------|--------|
| Healthy idle (0 connected, no errors) | *(hidden)* |
| 1+ connected | `MCP · playwright` (dim) |
| Needs OAuth | `MCP · auth: figma` (warning) |
| Recent failure | `MCP · fail: playwright` (error) |
| Mix | `MCP · fail: a · auth: b · c` (error color if any fail) |

Name lists are capped at 3 (`a, b, c +2`) so many servers stay readable.

## Motivation

`connected/configured` is an ops inventory metric, not a useful agent UI signal. The status bar should answer *"is there something I need to act on?"* / *"what's actively in use?"* — not *"how many sockets are open out of how many I declared?"*.

The detailed inventory remains available via `mcp({ })` / `/mcp` status (unchanged).

## Test plan

- [x] Unit tests for silence / connected / auth / fail / priority / cap / no-UI (`__tests__/update-status-bar.test.ts`, 9 cases)
- [x] Related suites still pass (`index-lifecycle`, `proxy-modes-manual-auth`, `init-elicitation`)
- [ ] Manual: start pi with only lazy servers → footer has no MCP line
- [ ] Manual: call a tool → footer shows `MCP · <server>` in dim
- [ ] Manual: OAuth-required server without credentials → `MCP · auth: <server>` in warning
- [ ] Manual: failed connect within backoff → `MCP · fail: <server>` in error
- [ ] Manual: idle timeout disconnects last server → status clears again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- Footer status no longer shows a permanent `MCP: 0/N servers` inventory ratio. Healthy idle (lazy, unused) is silent; active connections show server names in dim; `needs-auth` and recent failures surface as actionable `auth:` / `fail:` warnings. Fixes #93.
+
 ## [2.11.0] - 2026-07-03
 
 ### Changed
--- a/__tests__/update-status-bar.test.ts
+++ b/__tests__/update-status-bar.test.ts
@@ -1,0 +1,156 @@
+import { describe, expect, it, vi } from "vitest";
+import { updateStatusBar } from "../init.ts";
+import type { McpExtensionState } from "../state.ts";
+
+function makeUi() {
+  return {
+    setStatus: vi.fn(),
+    theme: {
+      fg: vi.fn((color: string, text: string) => `[${color}]${text}`),
+    },
+  };
+}
+
+function makeState(overrides: {
+  servers?: string[];
+  connections?: Record<string, { status: "connected" | "needs-auth" | "closed" }>;
+  failures?: Record<string, number>;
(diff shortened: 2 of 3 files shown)


Boubalou:
That was way too early, the idea is good but not ready nor tested by myself.

PR #194, opened 2026-07-20
fix: recover streamable HTTP session after server restart

damselem:
# fix: recover streamable HTTP session after server restart

## Problem

When a streamable HTTP MCP server restarts (deploy, crash, scale-up), it loses its in-memory sessions. The adapter keeps sending its old `Mcp-Session-Id`, the server answers 404, and every tool call fails from then on. The only way out today is restarting the whole Pi process, even though the server is healthy again.

The MCP spec already says what a client must do here:

> When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it MUST start a new session by sending a new `InitializeRequest`.

The adapter did not implement this. This PR does.

## What the fix does

When a tool call fails with the spec's exact "session is gone" signal, the adapter opens a fresh session and retries that call once. The model and the user never see the error, just a slightly slower call.

Four small pieces:

1. **Exact detection, no guessing** (`session-recovery.ts`):
   ```ts
   err instanceof StreamableHTTPError && err.code === 404 && hadSessionId
   ```
   No message-text matching. No 400 (ambiguous). No match for aborts, or for a 404 when no session was ever established (that is a config problem, not an expired session).

2. **One shared reconnect** (`server-manager.ts`, `reconnect()`): concurrent calls that hit the same dead session share a single reconnect instead of racing. It also never closes a connection that was already replaced by a newer one (identity check), so a slow failing call cannot tear down the fresh connection a faster call just built.

3. **One wrapper at every call site** (`withSessionRecovery`): used by the proxy tool, by direct tools, and by the embedded UI proxy. Retries exactly once. Any second failure goes through the existing error handling unchanged.

4. **Honest connection status**: `client.onclose` now marks the connection `"closed"` (with the same identity check). The SDK's internal `transport.onclose` is left alone on purpose, it owns pending-request cleanup.

## Why retrying is safe

The spec requires the server to reject an unknown session id before processing the request. A 404 therefore means the original call never ran. Retrying it once cannot double-execute anything.

## What this does not change

- Non-session errors behave exactly as before (tool `isError` results, auth errors, aborts, anything else).
- No retry loops, no backoff, no new config options. One reconnect, one retry, done.
- Known gap, on purpose: the manager's own `readResource()` (only used to fetch the UI host page) is not wired up, to keep the diff small. It self-heals on the next tool call.

## Tests

23 new tests across 4 files: the predicate table (what matches and what must not), transparent recovery through the real proxy, direct-tools, and UI-proxy paths, single-retry behavior, concurrency (one shared reconnect, no torn-down fresh connections), and the `onclose` identity guard. Four existing test files that mock the SDK's `streamableHttp.js` needed one added export (`StreamableHTTPError`) in their mock factories, nothing else changed in them.

Full suite green (two pre-existing failures on `main`, unrelated, missing example build artifacts). `tsc --noEmit` clean.

## Related upstream work

`modelcontextprotocol/typescript-sdk` PR #2125 proposes the same spec clause inside the SDK client. It is unmerged, and even merged it would only label the error, not retry. If it ships, this adapter code gets simpler (swap the predicate for the SDK's typed error), with no behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Surfaced MCP server `instructions` from the initialize handshake: captured at connect time, cached alongside tool metadata, shown as a truncated head in the `mcp` proxy tool description, previewed in `mcp({ server: "name" })` listings, and available in full via the new `mcp({ instructions: "name" })` mode. Thanks @JeongJuhyeon for issue #188 and PR #189.
 
 ### Fixed
+- Recovered Streamable HTTP MCP sessions after a server restart invalidates the previous session ID. Thanks @damselem for PR #194.
 - Used server-advertised OAuth protected-resource metadata during authorization so resource servers can point Pi at the correct authorization server. Thanks @jameswarren for issue #173 and PR #174.
 - Dropped inherited HTTP auth when a higher-precedence MCP config repoints a server URL, while preserving explicit OAuth disable flags. Thanks @ductiletoaster for PR #182.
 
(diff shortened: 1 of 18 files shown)


damselem:
@nicobailon sending this your way for review. Let me know if you have some feedback.

Issue #196, opened 2026-07-21
MCP connection failures show generic "failed" with no actionable cause


parkuman:
When an MCP server fails to connect, the UI only shows `failed` (panel) or `failed Ns ago` (`/mcp` status), and the underlying error is a generic `MCP error -32000: Connection closed`. There's no indication of *why* it failed, which makes debugging pretty opaque.

The example situation I ran into that inspired this was a stdio-based MCP (Loki) that runs a Docker container. This was simply saying `failed` in the MCP panel. Upon investigating, I didn't have Docker Desktop running! It would have been helpful to see this immediately in the panel so I didn't have to go looking.

**Before:**

https://github.com/user-attachments/assets/45ba707e-a62f-4666-808e-b7b922061d36

**Proposed Idea:**
Surface the error right in the panel on cursor. [PR here.](https://github.com/nicobailon/pi-mcp-adapter/pull/197)

https://github.com/user-attachments/assets/41ba4fe7-8af2-4010-98d3-1d1e6faa6094

parkuman:
Merged!

PR #197, opened 2026-07-21
feat: surface MCP connection failure reason

parkuman:
Persist the failure message and display it in the `/mcp` panel when that mcp server is under the cursor. This helps quickly diagnose issues with connecting to MCP servers rather than taking a guess at the failure reason. Users can use `ctrl+y` when the cursor is on the failed server to copy the error message for later use.

Captures stdio child stderr so failures show the real cause (e.g. `Cannot connect to the Docker daemon`) instead of a bare `MCP error -32000: Connection closed`. Addresses #196.

**Before**

https://github.com/user-attachments/assets/f0cf1850-804d-42a1-abf1-b122d7f91a65

**After**

https://github.com/user-attachments/assets/23b3626b-f68c-4840-a41f-d40f27ae57d2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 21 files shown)


parkuman:
Have added another commit that adds a shortcut to copy the error message:


https://github.com/user-attachments/assets/c5d55bca-ac08-43d3-ae18-4c81d93ec4ab

PR #220, opened 2026-07-25
fix: tolerate -32601 from tools/list so resources-only servers can connect

azarakovskiy:
Spec-correct MCP servers may omit any capability; a server that does not declare `tools` (or that declares it but does not implement `tools/list`) responds to a `tools/list` call with `-32601 Method not found`. 

The connect path in `McpServerManager` calls `fetchAllTools` in parallel with `fetchAllResources` / `fetchAllPrompts`; only the latter two catch that error, so any spec-correct resources-only server failed to connect with "MCP error -32601: tools not supported", and no resource-derived tool was ever registered.

**Fix:**
Wrap the body of `fetchAllTools` in the same try/catch used by `fetchAllResources`: re-throw on abort, otherwise return `[]`.

### DIFF
--- a/__tests__/fixtures/resources-only-server.mjs
+++ b/__tests__/fixtures/resources-only-server.mjs
@@ -1,0 +1,27 @@
+import { Server } from "@modelcontextprotocol/server";
+import { StdioServerTransport } from "@modelcontextprotocol/server/stdio";
+
+// A minimal MCP server that advertises resources only — no `tools`, no
+// `prompts`. Used by resources-capability.test.ts to check that the adapter
+// connects without aborting when tools/list is not supported, the spec-
+// canonical shape for read-only data sources.
+const server = new Server(
+  { name: "resources-only-server", version: "1.0.0" },
+  { capabilities: { resources: {} } },
+);
+
+const resource = {
+  name: "resource_name",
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks @azarakovskiy for the focused report and fixture.

I reproduced this against exact head `cbda01ff30f4c658b5a84490a36ff20c9b243e97` and current `main`. With the current `@modelcontextprotocol/client` v2 dependency, `Client.listTools()` already checks the negotiated server capabilities and returns an empty list without sending `tools/list` when `tools` is absent. The resources-only fixture therefore connects successfully on unchanged `main` and exposes its resource-derived tool.

The remaining failing case is the synthetic client mock, which bypasses that production capability check. The PR's focused suite also currently fails because the fixture defines `resource_name` while the assertion expects `current_price`. More importantly, the proposed catch-all would suppress transport, protocol, authorization, and malformed-response failures beyond the stated `-32601` case.

I'm closing this because there is no current production gap for a spec-compliant resources-only server. If a real server advertises `tools` but returns `-32601`, please share its initialization capabilities and trace; that would be a separate compatibility case we can narrow to the actual SDK error shape without hiding unrelated failures.

PR #226, opened 2026-07-27
fix(oauth): prefer configured scope over discovery

### LINKED ISSUE #225: Configured oauth.scope should override discovered WWW-Authenticate scope
## Summary

When an MCP server's auth discovery / `WWW-Authenticate` challenge advertises a scope, Pi's configured `oauth.scope` should take precedence over the discovered scope. Currently, the discovered scope can override the explicit config value during OAuth authorization-code flows.

This breaks Snowflake managed MCP servers in accounts where the discovered/default scope is rejected by account policy.

## Reproduction

Configure a Snowflake managed MCP server in `mcp.json` with an explicit role scope:

```json
{
  "mcpServers": {
    "snowflake": {
      "url": "https://<account>.<region>.aws.snowflakecomputing.com/api/v2/databases/<db>/schemas/<schema>/mcp-servers/<server>",
      "auth": "oauth",
      "oauth": {
        "clientId": "<client-id>",
        "clientSecret": "<client-secret>",
        "redirectUri": "http://localhost:3118/callback",
        "scope": "session:role:MCP_ROLE"
      }
    }
  }
}
```

Then start auth:

```text
/mcp-auth snowflake
```

## Expected behavior

The generated authorization URL should include the configured scope:

```text
scope=session%3Arole%3AMCP_ROLE
```

## Actual behavior

The generated authorization URL used the discovered scope from Snowflake instead:

```text
scope=session%3Arole%3Aall
```

Snowflake then rejected the OAuth login with an error like:

```text
The role ALL requested has been explicitly blocked for use with this application by an administrator.
```

This happens because Snowflake managed MCP advertises `session:role:all`, but many Snowflake OAuth integrations/accounts block privileged roles such as `ACCOUNTADMIN`, `ORGADMIN`, and `SECURITYADMIN`. Requesting `role:all` can therefore fail even when the user explicitly configured a valid least-privileged MCP role.

## Local workaround tested

I locally patched `mcp-auth-flow.ts` so configured scope wins after discovery:

```ts
runSdkAuth(authProvider, {
  serverUrl,
  ...discovery,
  ...(config.scope ? { scope: config.scope } : {})
})
```

instead of:

```ts
runSdkAuth(authProvider, {
  serverUrl,
  ...discovery
})
```

After this change, `/mcp-auth snowflake` generated the correct URL with:

```text
scope=session%3Arole%3AMCP_ROLE
```

and Snowflake OAuth succeeded.

## Why this seems like client behavior

A server-provided challenge is useful as a default, but an explicit user/client config value should be treated as an override. This is especially important for providers like Snowflake where the requested OAuth role is an account-policy-sensitive choice and users may need to select a specific least-privileged role.

adity982:
## Summary

- preserve OAuth discovery metadata while letting an explicit `oauth.scope` override the scope advertised by `WWW-Authenticate`
- carry the resolved scope through both authorization start and code completion
- add regression coverage for the configured-vs-discovered scope conflict

## Validation

- `npm test -- __tests__/mcp-auth-flow-client-credentials.test.ts` (30 tests passed)

Closes #225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Let configured `oauth.scope` override OAuth discovery scopes during authorization flows. Thanks @viggy28 for issue #225 and @adity982 for PR #226.
+
 ## [2.16.0] - 2026-07-30
 
 ### Added
--- a/__tests__/mcp-auth-flow-client-credentials.test.ts
+++ b/__tests__/mcp-auth-flow-client-credentials.test.ts
@@ -874,6 +874,34 @@ describe("mcp-auth-flow explicit auth", () => {
     expect(getOAuthState("direct-complete")).toBeUndefined();
   });
 
+  it("prefers a configured scope over discovery for start and completion", async () => {
+    mocks.fetch.mockResolvedValueOnce(new Response(null, {
+      headers: { "www-authenticate": 'Bearer scope="session:role:all"' },
+    }));
+    mocks.sdkAuth.mockImplementationOnce(async (provider) => {
+      await provider.redirectToAuthorization(new URL("https://auth.example.com/authorize"));
+      return "REDIRECT";
+    });
+    const { completeAuth, startAuth } = await import("../mcp-auth-flow.ts");
+
+    await startAuth("configured-scope", "https://api.example.com/mcp", {
+      url: "https://api.example.com/mcp",
+      auth: "oauth",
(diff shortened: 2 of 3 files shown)


nicobailon:
I reviewed this from fresh context against current `main` and found no blockers. Preferring the configured `oauth.scope` over the discovered scope is the behavior I want for #225, and the change validated cleanly when merged into current `main`.

It is still marked as a draft, so I am not merging it on your behalf. Mark it ready for review when you are happy with it and I will run CI and land it. If you would rather not carry it forward, say so and I will land the same change as a maintainer PR with credit to you. Thanks @adity982.

adity982:
Thanks for the fresh review and confirmation against current main. The focused scope and validation are complete, so I have marked this ready for review.

Issue #232, opened 2026-07-29
perf: collapsed MCP tool results full-wrap on every composer invalidate


hyknerf:
## Summary

Collapsed MCP tool-result rows in the Pi TUI wrap the **entire** result payload on every invalidate. Pi rebuilds custom tool rows on each composer keystroke (`ToolExecutionComponent.invalidate()` → `updateDisplay()` → `renderResult()`), so after large MCP dumps a long session makes the composer feel laggy.

In `tool-result-renderer.ts`, `renderMcpToolResult` builds a `CollapsibleText` that always holds the full themed body. On `render(width)`, collapsed mode previously did:

1. `new Text(fullBody).render(width)` — full ICU/grapheme wrap via `Intl.Segmenter`
2. slice to `maxCollapsedLines`
3. append `…` / Ctrl+O hint

So collapse only limited **display**, not **work**.

This is TUI-only. Model context / proxy token cost is unchanged.

### Live session

On a long `pi -c` session with many large MCP tool results:

- Pi process pegged ~100% CPU while typing in the composer
- `sample` stack: main thread → `uv_run` timers → V8 → **`Intl.Segmenter` / `SegmentIterator`**
- Session had many large tool rows (MCP content up to ~50–60 KiB)

### A/B (same long session)

| Run | Result |
|-----|--------|
| Stock `pi-mcp-adapter` (npm) | Composer lag |
| Local fix via `-e` | Noticeably snappier |
| `--no-extensions` without mcp-adapter (other packages still loaded) | Still snappy |

So stock collapsed MCP rendering is sufficient to explain the lag on MCP-heavy transcripts.

### Microbench (representative ~100 KiB first-line dump, 50 rebuild+render cycles)

| Path | Time (this host) |
|------|------------------|
| Prefix-wrap collapsed path | ~10–25 ms |
| Full ICU wrap of collapsed body | ~215–250 ms |

## Proposed fix

When collapsed:

- Only wrap a character-budget prefix large enough to fill a few visual lines:
  `width * (maxCollapsedLines + 1) * slack`
- Keep expanded / error / `details.error` paths on full text (by design)
- Pass the real `expanded` flag into `formatMcpToolResultLines` so multi-line dumps are logically truncated before join

Tradeoffs:

- Collapsed preview is approximate (same start of the dump; tail stays hidden either way)
- Expanded / error full-wrap remains expensive when the user opts in
[...]

PR #233, opened 2026-07-29
perf: avoid full ICU wrap for collapsed MCP tool results

hyknerf:
Issue: #232

### What

Collapsed MCP tool-result rows were ICU-wrapping the **entire** payload on every TUI invalidate. Pi rebuilds custom tool rows on each composer keystroke (`invalidate` → `updateDisplay` → `renderResult`), so large MCP dumps made typing lag in long sessions.

This changes `CollapsibleText` so collapsed mode only wraps a small character-budget prefix large enough to fill a few visual lines, then slices and shows the existing `…` / Ctrl+O chrome. Expanded and error paths still wrap the full body.

Also passes the real `expanded` flag into `formatMcpToolResultLines` so multi-line dumps are logically truncated before join (previously the render path always passed `true`).

### Why

- Collapse previously limited **display**, not **work**: `new Text(fullBody).render(width)` still ran `Intl.Segmenter` over multi-10KB dumps.
- Live `sample` on a lagging `pi -c` session showed main-thread time in `SegmentIterator` while typing.
- A/B on the same long session: stock adapter laggy; local fix snappy; no mcp-adapter (other extensions still loaded) also snappy.

Measured on a ~100 KiB first-line dump, 50 product-path rebuild+render cycles on this host:

| Path | Time |
|------|------|
| Prefix-wrap (this PR) | ~10–25 ms |
| Full ICU wrap | ~215–250 ms |

### Behavior notes

- Collapsed preview is approximate: same start of the dump, still capped to `maxCollapsedLines` visual rows. Tail remains hidden until Ctrl+O / error expand.
- Expanded, `isError`, and `details.error` paths are unchanged and still full-fidelity.
- Residual O(n) split/theme/join of a huge single line before the prefix slice remains; ICU wrap was the dominant cost.
- Model context / proxy token cost is unchanged (TUI-only).

### Tests

- Added a regression test that rebuilds via `renderMcpToolResult(...)` each iteration (matches Pi's product path), renders a multi-10KB payload collapsed, asserts the tail stays hidden, and keeps a time budget that full ICU wrap exceeds on this class of payload.
- Existing collapse / identity / error / expanded cases remain green.
- Locally: `npx tsc --noEmit`, `npx vitest run __tests__/tool-result-renderer.test.ts` (24 passed). No public API change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Collapsed MCP tool results no longer ICU-wrap the full payload on every TUI invalidate. Collapsed rows wrap only a small prefix, which stops composer keystroke lag after large MCP dumps.
+
 ## [2.15.0] - 2026-07-25
 
 ### Added
--- a/__tests__/tool-result-renderer.test.ts
+++ b/__tests__/tool-result-renderer.test.ts
@@ -159,6 +159,27 @@ describe("MCP tool result renderer", () => {
     expect(output).not.toContain("segment-8");
   });
 
+  it("keeps collapsed rendering cheap for multi-10KB MCP dumps", () => {
+    const huge = `${"word ".repeat(20_000)}\n${"line\n".repeat(2_000)}tail-marker`;
+    const toolResult = result([{ type: "text", text: huge }]);
+
+    // Match the product path: each TUI invalidate rebuilds via renderResult().
+    const started = performance.now();
+    let component = renderMcpToolResult(toolResult, collapsedOptions, plainTheme, { isError: false });
+    for (let i = 0; i < 50; i++) {
+      component = renderMcpToolResult(toolResult, collapsedOptions, plainTheme, { isError: false });
+      component.render(80);
+    }
+    const elapsedMs = performance.now() - started;
+
+    const output = component.render(80).join("\n");
+    expect(output).toContain("word");
+    expect(output).toContain("Ctrl+O to expand");
+    expect(output).not.toContain("tail-marker");
+    // Full-text ICU wrap of this payload is hundreds of ms; prefix wrap stays small.
+    expect(elapsedMs).toBeLessThan(100);
+  });
+
   it("shows proxy call result identity without hiding the third content line", () => {
     const output = renderMcpToolResult(
       result([{ type: "text", text: "one\ntwo\nthree\nfour" }], { mode: "call", server: "figma", tool: "get_nodes" }),
--- a/tool-result-renderer.ts
+++ b/tool-result-renderer.ts
@@ -34,6 +34,12 @@ export interface McpToolResultDisplay {
 
 const DEFAULT_MAX_CALL_INPUT_CHARS = 1500;
 const DEFAULT_MAX_COLLAPSED_LINES = 3;
+/**
+ * Collapsed TUI rows only need enough source text to fill a few visual lines.
+ * Capping here keeps ICU/grapheme wrapping off multi-10KB MCP dumps on every
+ * keystroke invalidate (Pi rebuilds tool rows on each composer render).
+ */
+const COLLAPSED_RENDER_CHAR_SLACK = 8;
 
 class CollapsibleText implements Component {
   constructor(
@@ -45,8 +51,22 @@ class CollapsibleText implements Component {
   ) {}
 
   render(width: number): string[] {
-    const lines = new Text(this.text, 0, 0).render(width);
-    if (this.expanded || lines.length <= this.maxCollapsedLines) return lines;
+    if (this.expanded) {
+      return new Text(this.text, 0, 0).render(width);
+    }
+
+    // Only wrap a prefix large enough to fill the collapsed viewport. Full-text
+    // wrap of large MCP results dominates the main-thread sample (Intl.Segmenter).
+    // Multi-line dumps are already truncated by formatMcpToolResultLines.
+    const safeWidth = Math.max(1, Math.floor(width));
+    const charBudget = safeWidth * (this.maxCollapsedLines + 1) * COLLAPSED_RENDER_CHAR_SLACK;
+    const prefix = this.text.length > charBudget
+      ? this.text.slice(0, charBudget)
+      : this.text;
+    const fullyIncluded = prefix === this.text;
+
+    const lines = new Text(prefix, 0, 0).render(width);
+    if (fullyIncluded && lines.length <= this.maxCollapsedLines) return lines;
 
     return [
       ...lines.slice(0, this.maxCollapsedLines),
@@ -191,7 +211,9 @@ export function renderMcpToolResult(
 
   const hasErrorDetails = Boolean(result.details.error);
   const expanded = options.expanded || context?.isError === true || hasErrorDetails;
-  const display = formatMcpToolResultLines(result, true);
+  // Keep logical collapse for multi-line dumps; CollapsibleText still handles
+  // single-line visual wrap without scanning the full body when collapsed.
+  const display = formatMcpToolResultLines(result, expanded);
   const identity = formatMcpToolResultIdentity(result.details);
   const output = [
     ...(identity ? [activeTheme.fg("muted", identity)] : []),


hyknerf:
## CPU profile follow-up (before/after)

Compared two Node CPU profiles of the same workflow:

```bash
cd <project>
node --cpu-prof --cpu-prof-dir=./profiles --cpu-prof-name=continue.cpuprofile \
  <pi-repo>/packages/coding-agent/dist/cli.js -c
```

| Profile | Build |
|--------|--------|
| **before** | `nicobailon/pi-mcp-adapter` (unpatched) |
| **after** | this branch (`fix/collapsed-mcp-result-render-lag` / `hyknerf` fork) |

**Note:** after has a longer wall clock (~28s → ~68s) because the session stayed open longer. ~91% of the after profile is `(idle)`. Compare active CPU, not wall time.

### Active CPU

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| tui package self | 20007 ms | 3849 ms | **-81%** |
| tui utils self | 19516 ms | 3342 ms | **-83%** |
| `visibleWidth` self | 10596 ms | 181 ms | **-98%** |
| `breakLongWord` self | 1919 ms | 171 ms | **-91%** |
| `splitIntoTokensWithAnsi` self | 4361 ms | 2484 ms | **-43%** |
| renderer on stack (`tool-result-renderer`) | 19867 ms | 2996 ms | **-85%** |
| wrap path via renderer | 19572 ms | 2979 ms | **-85%** |
| GC | 455 ms | 116 ms | **-74%** |

### Hot path

**Before** (dominated sample):

```
tui doRender
  → tool-execution.render
    → tool-result-renderer
      → Text.render → wrapTextWithAnsi / wrapSingleLine
        → visibleWidth (~37% self)
        → splitIntoTokensWithAnsi (~15%)
        → breakLongWord (~7%)
```

**After:** same stack still appears, but ~2.5s vs ~20s. `visibleWidth` / grapheme work is effectively gone on collapsed rows, which matches the prefix char-budget change.

### Residual (optional, not blocking)

1. `splitIntoTokensWithAnsi` is the main remaining non-idle cost (~3.6% of the long after profile; larger share of active CPU).
2. Expanded / error results still full-wrap by design — huge expanded dumps can still lag.
3. A TUI-level wrap/width cache in pi would help all tool renders, not only MCP.

### Verdict

The collapsed-prefix fix removes on the order of **16–17s** of renderer/wrap CPU on this session and eliminates the `visibleWidth` hotspot. Remaining cost is small relative to before; continue/session load itself was never the bottleneck.

This comment is AI-generated from local `continue.cpuprofile` / `continue2.cpuprofile` analysis.

chapmanb:
Independent reproduction of #232, plus **a suggested addition to this branch** that I think is worth folding in before merge.

**tl;dr** — there is a second invalidate trigger that this PR doesn't cover: Pi's TUI re-renders every component on an **80 ms timer while the agent is working**, with no state change and no `invalidate()`. Because `CollapsibleText` constructs its `Text` *inside* `render()`, that cache is always cold, so a completely no-op frame still pays a full wrap. On my transcripts this branch lands at **~21% of a core**, and holding the `Text` instance takes it to **~0.1%**. Three-line change, output byte-identical.

## Suggested addition: hold the `Text` instead of rebuilding it in `render()`

```diff
 class CollapsibleText implements Component {
   constructor(...) {}

+  private fullComponent: Text | undefined;
+  private prefixComponent: Text | undefined;
+  private prefixWidth: number | undefined;
+
   render(width: number): string[] {
     if (this.expanded) {
-      return new Text(this.text, 0, 0).render(width);
+      this.fullComponent ??= new Text(this.text, 0, 0);
+      return this.fullComponent.render(width);
     }

     const safeWidth = Math.max(1, Math.floor(width));
     const charBudget = safeWidth * (this.maxCollapsedLines + 1) * COLLAPSED_RENDER_CHAR_SLACK;
     const prefix = this.text.length > charBudget ? this.text.slice(0, charBudget) : this.text;
     const fullyIncluded = prefix === this.text;

-    const lines = new Text(prefix, 0, 0).render(width);
+    if (!this.prefixComponent || this.prefixWidth !== width) {
+      this.prefixComponent = new Text(prefix, 0, 0);
+      this.prefixWidth = width;
+    }
+    const lines = this.prefixComponent.render(width);
     if (fullyIncluded && lines.length <= this.maxCollapsedLines) return lines;
```

`Text` already caches on `(text, width)` — constructing a new one per `render()` just guarantees the cache never hits. Fields are lazy to avoid any class-field-vs-parameter-property initialisation-order dependency. The `prefixWidth` check is needed because `prefix` itself depends on `width`; I verified 80 → 200 → 80 resize round-trips produce identical output.

This is complementary to the prefix wrap, not a replacement — they fix disjoint triggers:

| trigger | mechanism | fixed by |
|---|---|---|
| composer keystroke | `ToolExecutionComponent.invalidate()` → `updateDisplay()` → `renderMcpToolResult()` builds a **new** `CollapsibleText` | prefix wrap (this PR) |
| **spinner timer, no user input** | `doRender()` → `Container.render()` → `.render(width)` on the **same** instance | held `Text` (the above) |

An instance-level cache can't help the first (the instance is discarded), and the prefix wrap can't fully help the second (it still re-wraps the budgeted prefix every frame).

## The second trigger, and why it's the worse one

#232 and the changelog both attribute the invalidate to the composer. That's real, but I hit this without typing at all.

I profiled a live Pi session sitting at **98% CPU for 68 minutes** during an unattended multi-step agent run. Call tree:

```
processTimers → listOnTimeout
  → pi-tui/dist/tui.js:537                      (render timer)
    → doRender  tui.js:980
      → Container.render  tui.js:96
        → ToolExecutionComponent.render          tool-execution.js:177
          → Box.render
            → pi-mcp-adapter/tool-result-renderer.ts:47   ← 31.5% of ALL samples
              → Text.render → wrapTextWithAnsi → wrapSingleLine
                  → visibleWidth            (17.8% self)
                  → splitIntoTokensWithAnsi (6.4%)
                  → graphemeWidth           (3.4%)
```

**94% of all non-idle time** in that one path. (The capture is 64.6% idle only because the agent finished its turn partway through — CPU fell 98% → 1% the instant it did, which is itself the tell.)

The driver is `Loader`'s spinner: `setInterval(…, 80ms)` at `pi-tui/dist/components/loader.js:54`, so 12.5 renders/sec baseline and up to ~62/sec at the `MIN_RENDER_INTERVAL_MS = 16` floor in `tui.js:123` while streaming. Whenever the agent is working, that timer requests a render — and every collapsed MCP row in the transcript pays a full wrap. The cost is cumulative and permanent for the life of the process.

## Measurements

Accumulated transcript of **5 single-line + 6 multi-line** MCP rows (~43–51 KB each, matching the distribution in my real sessions), themed, width 120. Timer path: repeated `.render()` on the same components, no state change.

| variant | ms/frame | % of one core @12.5 fps | @62 fps |
|---|---|---|---|
| upstream 2.15.0 | 331.96 | 415% | 2058% |
| `expanded` flag only | 164.53 | 206% | 1020% |
| **PR #233 as-is** | **16.73** | **21%** | **104%** |
| **PR #233 + held `Text`** | **0.053** | **0.1%** | **0.3%** |

Per-row, single measurements:

| payload | upstream | flag only | PR #233 | + held `Text` |
|---|---|---|---|---|
| multi-line, 550 lines, 45 KB | 45.71 ms | 0.032 ms | 0.031 ms | ~0.005 ms |
| single line (minified JSON), 43 KB | 33.86 ms | 35.98 ms | 3.71 ms | ~0.005 ms |

Node v22.20.0, Pi 0.83.0, 8-core Linux, mean of 15–30 iterations after warmup.

Three things worth pulling out:

**1. `plainTheme` understates the multi-line case by ~55×** (0.833 ms → 45.71 ms). `renderMcpToolResult` applies `activeTheme.fg("toolOutput", line)` per line, so in the real TUI the string reaching `wrapTextWithAnsi` is dense with SGR codes and `visibleWidth` / `splitIntoTokensWithAnsi` get much more expensive. The regression test added here uses `plainTheme`, so it exercises the cheap path — a themed variant would be a stronger guard. At 45 ms/frame a *single* row already exceeds a 62 fps budget, which is what turns this from lag into saturation.

**2. Both halves of the current PR are load-bearing, on disjoint payload shapes.** The `expanded` flag does essentially all the work for multi-line dumps (45.71 → 0.03 ms) and *nothing* for single-line ones (33.86 → 35.98 ms). The prefix wrap does essentially all the work for single-line dumps (33.86 → 3.71 ms). Worth noting in case anyone is tempted to take only the one-token change — I initially did exactly that locally and it left me at 206% of a core.

**3. Single-line payloads are common.** Across my 25 most recent sessions, 5 of 11 large MCP text blocks were a single line, up to 51,198 chars — minified JSON from servers that don't pretty-print. I'd assumed multi-line was the norm and was wrong.

## Review notes on the existing diff

I went looking for problems with the prefix slice and didn't find any:

- **ANSI safety.** My first worry was that `this.text.slice(0, charBudget)` could cut an SGR sequence in half, since the body is themed per line. It doesn't, and `COLLAPSED_RENDER_CHAR_SLACK = 8` is what saves it — any damage lands ~8× beyond the visible window and is discarded by `lines.slice(0, this.maxCollapsedLines)`. Separately, once the `expanded` fix is in, collapsed multi-line text is small enough that the slice never fires, and the single-line case has one SGR pair around the whole line so the cut lands in plain content.
- **Output equivalence.** Byte-identical to 2.15.0 (raw, ANSI included) across 24 scenarios — 12 payload shapes × plain and themed — covering under/at/over the collapse threshold, single huge line, one long line plus short ones, wrapping-length lines, missing identity, expanded, both error paths (`context.isError` and `details.error`), and empty results. That holds with the held-`Text` change applied, including on the second render.
- I did notice an SGR set/reset imbalance in collapsed output for long wrapped lines, but it's present identically in 2.15.0, so it's pre-existing `pi-tui` wrapping behaviour and not something this PR introduces.

## Profiling an already-running process

Slightly different from the `--cpu-prof` restart in the PR description, and useful when the pathological state took an hour to accumulate. It also works under `kernel.yama.ptrace_scope=1`, where `gdb -p` / `strace -p` can't attach, because Yama governs `ptrace(2)` and not `kill(2)`:

```bash
kill -USR1 <pi-pid>                          # opens the V8 inspector on 127.0.0.1:9229
curl -s http://127.0.0.1:9229/json/list      # confirm the target
# then drive CDP Profiler.start / Profiler.stop over the websocket
```

Caveats: Node writes `Debugger listening on…` to stderr, which corrupts the TUI until `Ctrl+Z` + `fg` forces a full redraw, and the inspector stays open for the life of the process.

---

Happy to open the held-`Text` change as a PR against your branch, or you're welcome to just take the diff above. I can also share the standalone benchmark if it'd be useful as a fixture. Thanks for chasing this down — the diagnosis in #232 was right, I just came at it from the timer side.

nicobailon:
Landed as #243 (squashed to 8f593aaacad3d30abe263a4b6ab61e62dadb5a29).

I rebased your change onto current `main` and folded in @chapmanb's held-`Text` follow-up, which is what keeps the repeated-render path from redoing the wrap. Closing this as superseded rather than merged only because of the rebase; the approach and the credit are yours. Thanks @hyknerf.

PR #234, opened 2026-07-29
fix: double connection close race

golota60:
`client.close()` should already close `transport` underneath

This double-close race when running `pi` with `--mode json` to sometimes not shut down gracefully, leading to a hanging `pi`, awaiting a close which will never happen(cause one `.close()` will never resolve)


golota60:
needs adjustment; will open a separate PR after double-checking everything

golota60:
Re-opened in https://github.com/nicobailon/pi-mcp-adapter/pull/235

PR #243, opened 2026-07-30
perf: avoid repeated collapsed result wrapping

nicobailon:
Maintainer replacement for #233, rebased onto current `main`.

Collapsed MCP tool results were re-wrapping the entire payload through the ICU wrapper on every TUI render, so a large result made composer keystrokes lag. This caches the collapsed rendering work across repeated renders and bounds the payload before wrapping, since collapsed mode only ever shows a short prefix.

Credit:
- Thanks @hyknerf for #233 and issue #232. The approach here is yours.
- Thanks @chapmanb for the held-`Text` follow-up, which is what keeps the repeated-render path from redoing the work.

Validation:
- `npx vitest run __tests__/tool-result-renderer.test.ts` (24 tests)
- `npx tsc --noEmit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Prevented `/mcp` and `/mcp-auth` from crashing when the OS OAuth credential store is unavailable.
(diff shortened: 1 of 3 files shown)


chapmanb:
Thanks much for the merge and release, it's working great now. Really appreciate your time creating and maintaining this package, we use a ton of MCP servers and it's been nice to be able to leverage them with pi.

PR #280, opened 2026-08-04
fix: exec absolute-path $BROWSER directly on darwin instead of via open -a

kipz:
## Summary

- On darwin, `execOpen` always shells out to the bare `open` command, even when a `browser` override is supplied — the override only ever gets passed as `open -a <browser>`, which expects an application name/bundle rather than an executable path.
- Sandboxing/launcher tools (e.g. [nono](https://github.com/nolabs-ai/nono)) that set `$BROWSER` to an absolute binary path (e.g. `/usr/bin/open`) have no way to make use of it on macOS, unlike the existing Linux/Windows code paths where the override already replaces the executable.
- This changes the darwin branch so that when `browser` is an absolute path, it's exec'd directly; a bare app name still goes through `open -a` as before.

## Test plan

- [x] `npm run typecheck`
- [x] `npm test` (909/911 passing — the 2 pre-existing failures in `interactive-visualizer-server.test.ts` are unrelated, missing built example assets, and fail identically on `main`)
- [x] Added `__tests__/utils-exec-open.test.ts` covering: absolute-path browser exec's directly, non-path browser name still uses `open -a`, no browser falls back to bare `open`

### DIFF
--- a/__tests__/utils-exec-open.test.ts
+++ b/__tests__/utils-exec-open.test.ts
@@ -1,0 +1,44 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { platform } from "node:os";
+import { openUrl } from "../utils.ts";
+import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
+
+vi.mock("node:os", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:os")>();
+  return { ...actual, platform: vi.fn() };
+});
+
+function fakePi(): { pi: ExtensionAPI; exec: ReturnType<typeof vi.fn> } {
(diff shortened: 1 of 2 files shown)


kipz:
Thanks for the review.

@greptile-apps — good catch, the absolute-path branch's `signal` forwarding wasn't covered. Added a test with a real `AbortController` in 524a842 confirming it's passed through.

@chatgpt-codex-connector — no actionable findings, nothing to address there.

nicobailon:
Closing this as superseded by the maintainer replacement in #281, which has landed on `main`.

I used this PR as the source for the macOS `$BROWSER` fix, kept the direct absolute-executable behavior, and added the review fix so absolute `.app` bundle paths still go through `open -a`. The merge body credits you for #280.

Thanks for the fix.

PR #294, opened 2026-08-05
fix: forward OAuth callback issuer to SDK

### LINKED ISSUE #293: Bug: RFC 9207 `iss` is dropped before OAuth code exchange
While testing the extension against Cloudflare MCP
(`https://mcp.cloudflare.com/mcp`), I found that OAuth completion fails when
Cloudflare includes the RFC 9207 `iss` callback parameter.

`auth-complete` parses and validates `iss`, but does not pass it to the MCP
SDK's `auth()` call. The SDK then rejects the authorization-code exchange.

### Reproduction

1. Configure Cloudflare MCP:

   ```json
   {
     "mcpServers": {
       "cloudflare": {
         "url": "https://mcp.cloudflare.com/mcp"
       }
     }
   }
   ```

2. Run `mcp({ action: "auth-start", server: "cloudflare" })`.
3. Approve in the browser.
4. Complete with the full localhost callback URL, including `code`, `state`,
   and `iss`.

### Actual result

OAuth completion fails:

```text
Issuer mismatch in authorization response (RFC 9207):
expected "https://mcp.cloudflare.com", received undefined
```

The pending flow is then cleared, requiring authorization to be restarted.

### Cause

`completeAuth()` correctly parses `{ code, iss }` and validates `iss` locally,
but invokes the SDK with only:

```ts
runSdkAuth(pendingAuth.authProvider, {
  serverUrl: pendingAuth.serverUrl,
  authorizationCode: code,
  ...pendingAuth.discovery,
})
```

The SDK performs its own RFC 9207 validation and does not receive `iss`.

### Proposed fix

Pass the parsed issuer through:

```ts
runSdkAuth(pendingAuth.authProvider, {
  serverUrl: pendingAuth.serverUrl,
  authorizationCode: code,
  iss,
  ...pendingAuth.discovery,
})
```

### Context

PR #210 states that callback `iss` is propagated, but current `main` drops it
at the SDK boundary.

tkoenig:
## Summary

Restores forwarding of the RFC 9207 callback `iss` parameter to the SDK during manual OAuth completion.

PR #210 included the handoff. It was removed during the SDK v2 → v1 compatibility rollback in #242, while the v1 SDK still validates the authorization-response issuer.

Closes #293.

## Verification

- `npx vitest run __tests__/mcp-auth-flow-client-credentials.test.ts`
- `npm run typecheck`
- `npm run test:oauth`

### DIFF
--- a/__tests__/mcp-auth-flow-client-credentials.test.ts
+++ b/__tests__/mcp-auth-flow-client-credentials.test.ts
@@ -141,6 +141,13 @@ describe("mcp-auth-flow explicit auth", () => {
       `code=auth-code&state=${oauthState}&iss=${encodeURIComponent("https://auth.example.com")}`,
     )).resolves.toBe("authenticated");
     expect(mocks.sdkAuth).toHaveBeenCalledTimes(2);
+    expect(mocks.sdkAuth).toHaveBeenLastCalledWith(
+      expect.anything(),
+      expect.objectContaining({
+        authorizationCode: "auth-code",
+        iss: "https://auth.example.com",
+      }),
+    );
     expect(hasPendingAuth("rfc9207-missing")).toBe(false);
   });
 
--- a/mcp-auth-flow.ts
+++ b/mcp-auth-flow.ts
@@ -606,6 +606,7 @@ export async function completeAuth(
     const result = await abortable(runSdkAuth(pendingAuth.authProvider, {
       serverUrl: pendingAuth.serverUrl,
       authorizationCode: code,
+      ...(iss !== undefined ? { iss } : {}),
       ...pendingAuth.discovery,
     }), signal)
     throwIfAborted(signal)


ugur-murat-alt:
Tested this fix independently against the live Cloudflare MCP endpoint using `pi-mcp-adapter@2.20.1`, Pi `0.83.0`, and Node `24.18.0` on Linux.

**Before:** OAuth completion reproduced the RFC 9207 failure where the adapter had parsed/validated the callback issuer but the SDK reported that the authorization-response issuer was undefined.

**After applying the issuer handoff:**

- reloaded the patched adapter in the same live Pi runtime;
- completed a fresh OAuth flow in that same runtime;
- Cloudflare MCP reconnected with 3 tools / 0 resources;
- ran one read-only `search` against the public OpenAPI spec successfully;
- did not call `execute` or any mutation tool;
- kept adapter-side issuer validation enabled.

The installed SDK declarations also expose `AuthOptions.iss?: string`, and the changed source was clean under the TypeScript LSP. The conditional form in this PR is a good fit: it forwards `iss` when present without adding an explicit undefined option for non-RFC-9207 flows.

This provides an independent real-endpoint confirmation in addition to the regression assertion already included here.

ugur-murat-alt (approved):
Independent live Cloudflare MCP verification completed against pi-mcp-adapter 2.20.1: the RFC 9207 failure reproduced before the change, same-runtime OAuth and a public-spec read-only search passed after forwarding iss, and issuer validation remained enabled. No blockers found.

PR #346, opened 2026-08-13
fix: preserve valid direct-tool prefix characters

### LINKED ISSUE #342: Preserve provider-valid characters in direct-tool server prefixes
## Background

Issue #334 reported that configured MCP server names containing spaces could produce invalid direct-tool names such as `Bright Data_search_engine`, which OpenAI-compatible providers reject.

PR #338 fixed this by hex-escaping every non-alphanumeric character in server-name prefixes:

```text
"my server" → my_20_server
"my_server" → my_5f_server
```

## Problem

Some characters escaped by the current implementation—including underscores and hyphens—are already valid in provider function and tool names.

For example, configured servers named `my_server` and `my-server` expose:

```text
my_5f_server_get
my_2d_server_get
```

rather than preserving their valid characters:

```text
my_server_get
my-server_get
```

These original server names already satisfy the relevant character constraints, so escaping their valid characters appears unnecessary. It also makes direct tools harder to recognize and produces surprising names for existing configurations.

Escaping remains necessary for characters that providers do not accept. Any adjustment must also account for normalized-name collisions and provider identifier-length limits without routing a call to the wrong server or silently dropping a tool.

## Possible consideration

It may be possible to preserve provider-safe characters in emitted direct-tool names while applying normalization or disambiguation only when needed. Direct-tool registrations already retain the configured server and original tool identity, which might help avoid relying on the emitted name for dispatch.

This is only a possible direction. The appropriate naming, compatibility, collision-handling, and scope decisions are left to the maintainers.

## Expected outcome

- Characters accepted by provider tool-name constraints are preserved in emitted direct-tool names.
- For example, `my_server` and `my-server` produce names such as `my_server_get` and `my-server_get`, rather than hex-escaping `_` and `-`.
- Server names containing unsupported characters still produce provider-valid direct-tool identifiers.
- Distinct configured servers and tools cannot be confused after normalization.
- Regression coverage includes provider-valid characters, unsupported characters, and overlapping normalized names.

### LINKED ISSUE #343: Bug: Codebase Memory tools are renamed with `_2d_` fragments
## What is happening

The latest `pi-mcp-adapter` release changes Codebase Memory MCP tool names by inserting `_2d_` fragments into the generated server prefix.

Codebase Memory upstream repo:

https://github.com/DeusData/codebase-memory-mcp

## Affected MCP server

Codebase Memory / `codebase-memory-mcp`.

## Observed tool names

The adapter now exposes Codebase Memory tools with `_2d_` in the server-derived prefix, for example:

```text
codebase_2d_memory_2d_mcp_index_repository
codebase_2d_memory_2d_mcp_search_graph
codebase_2d_memory_2d_mcp_query_graph
codebase_2d_memory_2d_mcp_trace_path
codebase_2d_memory_2d_mcp_get_code_snippet
codebase_2d_memory_2d_mcp_search_code
codebase_2d_memory_2d_mcp_get_architecture
```

## Previous naming

Before the latest release, the Codebase Memory server prefix used underscore-normalized names derived from `codebase-memory-mcp`, e.g. names in the form:

```text
codebase_memory_mcp_<tool>
```

## Regression note

This appears after the server-prefix escaping change for #334. Because the Codebase Memory server name contains hyphens, the adapter now renders each hyphen as `_2d_` in the visible MCP tool names.

nicobailon:
Fixes #342.
Fixes #343.

This updates direct-tool server-prefix formatting so provider-valid `_` and `-` characters stay recognizable in emitted tool names, while unsupported characters still use deterministic escaping. It also keeps the existing duplicate warning-and-skip safety and fixes proxy lookup so an exact preserved hyphen name cannot route to an underscore server through the normalized fallback.

Examples now stay readable:
- `my_server_get`
- `my-server_get`
- `codebase-memory-mcp_<tool>`

Validation:
- `npx vitest run __tests__/abort-signal.test.ts __tests__/direct-tools.test.ts __tests__/resolve-server-from-tool-name.test.ts`
- `npm run typecheck`
- fresh review on `5d246f3`: found and proved the wrong-server proxy route
- retained writer fix at `cb96af5`
- follow-up review on `cb96af5`: no issues found
- rebase and changelog validation on `6597151`: focused tests and typecheck passed
- exact-head review on `6597151`: no issues found

Credit is included in `CHANGELOG.md` for @mightymatth and @JasonLandbridge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Kept `mcpScript` tool-call replies flowing when Pi runs as a Bun-compiled binary by starting worker execution without module-level top-level await. Thanks @AndreiKopylov for issue #340 and @thesobercoder for the verified fix.
+- Fixed direct-tool server prefixes to preserve provider-valid underscores and hyphens, including `codebase-memory-mcp`, while retaining safe escaping, collision handling, and exact proxy routing. Thanks @mightymatth for issue #342 and @JasonLandbridge for issue #343.
 
 ## [2.24.0] - 2026-08-13
 
--- a/__tests__/abort-signal.test.ts
+++ b/__tests__/abort-signal.test.ts
@@ -73,6 +73,43 @@ describe("AbortSignal propagation", () => {
     expect(state.manager.decrementInFlight).toHaveBeenCalledWith("demo");
   });
 
+  it("prefers an exact proxy tool name over an earlier normalized match", async () => {
+    const underscoreCallTool = vi.fn(async () => ({ content: [{ type: "text", text: "underscore" }] }));
+    const hyphenCallTool = vi.fn(async () => ({ content: [{ type: "text", text: "hyphen" }] }));
+    const state = {
+      config: {
+        settings: { toolPrefix: "server" },
+        mcpServers: {
+          my_server: { command: "underscore" },
+          "my-server": { command: "hyphen" },
+        },
+      },
+      manager: {
+        getConnection: vi.fn((server: string) => ({
+          status: "connected",
+          client: server === "my_server" ? { callTool: underscoreCallTool } : { callTool: hyphenCallTool },
+        })),
+        touch: vi.fn(),
+        incrementInFlight: vi.fn(),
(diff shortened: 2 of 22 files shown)


JasonLandbridge:
@nicobailon Thanks for the great plugin and quick fix!

PR #365, opened 2026-08-17
fix: don't hang /mcp panels in rpc, json and print modes

shixin-guo:
## Problem

`/mcp` and `/mcp setup` hang forever — no output, no response — for any host that embeds the adapter through `pi --mode rpc`.

`openMcpPanel`, `openMcpSetup` and `openMcpAuthPanel` guard on `ctx.hasUI` before showing their overlay. But pi documents `hasUI` as:

> Whether dialog-capable UI is available (**true in TUI and RPC modes**)

So the guard passes outside the terminal UI, and each function then awaits a panel like this:

```ts
await new Promise<void>((resolve) => {
  ctx.ui.custom((tui, _theme, keybindings, done) => createMcpPanel(..., () => {
    done(undefined);
    resolve();
  }), { overlay: true, ... });
});
```

Outside the TUI, `ctx.ui.custom()` is a headless stub — pi's `modes/rpc/rpc-mode.js` implements it as:

```js
async custom() {
    // Custom UI not supported in RPC mode
    return undefined;
},
```

The factory is never invoked, so `done` and `resolve` never fire and the promise never settles. The command handler stays blocked indefinitely.

### Reproduction

Against pi 0.83.0, driving the adapter over stdio RPC:

| command | before | after |
|---|---|---|
| `/mcp` | **0 frames, no response** (hangs) | server status text + `response success` |
| `/mcp status` | **0 frames, no response** (hangs) | server status text + `response success` |
| `/mcp tools` | works (uses `ctx.ui.notify`) | unchanged |

`/mcp tools` already worked, which is what pointed at `ctx.ui.custom()` as the specific culprit rather than the command plumbing.

## Fix

Guard overlay panels with the mode pi documents for exactly this purpose:

> `mode: ExtensionMode` — Current run mode. **Use `"tui"` to guard terminal-only UI such as custom components.**

...and fall back to output that already exists in this file rather than adding new surface area:

- `/mcp` and `/mcp status` → render the text status via the existing `showStatus()`
- `/mcp setup` and the auth panel → explain where the panel is available and point at the non-interactive alternative

`hasUI` checks are left in place, so the `!hasUI` behaviour is unchanged.

## Tests

Added `__tests__/commands-panel-non-tui.test.ts`. The stub `custom()` in it never calls back, mirroring the real RPC stub, so a regression shows up as the test's own hang-detector firing rather than a vague assertion failure. Reverting the `commands.ts` change fails 3 of its 4 cases, one with `openMcpSetup hung waiting for a custom UI panel`.

Three existing fixtures that exercise the overlay path only set `hasUI: true`; they now also set `mode: "tui"` to match what pi actually passes.

`npm run typecheck` is clean. Full suite is unchanged relative to `main`: the only failures are the pre-existing `interactive-visualizer-server` and `request-headers-command` ones, which pass in isolation and flake by ±1 under full-suite load on both `main` and this branch.

## Note on scope

This fixes the hang. It does not make the panels *render* outside a terminal — a host that wants the real interactive panel still has to implement `ctx.ui.custom()` itself. Degrading to text seemed like the right default here, since it uses renderers this file already has and matches how `/mcp tools` and `/mcp prompts` already behave.

### DIFF
--- a/__tests__/commands-auth.test.ts
+++ b/__tests__/commands-auth.test.ts
@@ -27,7 +27,7 @@ describe("authenticateServer", () => {
     await openMcpAuthPanel({
       programmaticConfig: false,
       config: { mcpServers: { disabled: { url: "https://example.test/mcp", auth: "oauth", disabled: true } } },
-    } as any, { getFlag: vi.fn() } as any, { hasUI: true, ui } as any);
+    } as any, { getFlag: vi.fn() } as any, { hasUI: true, mode: "tui", ui } as any);
 
     expect(ui.notify).toHaveBeenCalledWith("No OAuth-capable MCP servers are configured.", "warning");
     expect(ui.custom).not.toHaveBeenCalled();
@@ -44,7 +44,7 @@ describe("authenticateServer", () => {
       const definition = { url: "${MCP_AUTH_URL}", auth: "oauth" as const };
       const result = await authenticateServer("sentry", {
         mcpServers: { sentry: definition },
-      }, { hasUI: true, ui } as any);
+      }, { hasUI: true, mode: "tui", ui } as any);
 
       expect(result.ok).toBe(true);
       expect(mocks.authenticate).toHaveBeenCalledWith(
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks for the fix. I landed it through #373 with the same reviewed panel changes plus the required changelog credit for your contribution.

Closing this PR as superseded by #373.

PR #370, opened 2026-08-18
fix: converge stale keep-alive tool catalogs

### LINKED ISSUE #369: Keep-alive connections can stay stale after a remote MCP server is replaced
## Problem

We run a long-lived Pi session against a remote Streamable HTTP MCP server with `lifecycle: "keep-alive"` and direct tools enabled.

When the MCP server is restarted or replaced, its tool catalog can change and the old notification stream can disappear without the adapter marking the connection disconnected. The keep-alive check still sees `connected`, so it does not reconnect or run a fresh `tools/list`.

The existing Pi session then keeps a stale tool surface: newly added tools are missing and removed tools can remain visible. Restarting Pi or explicitly reconnecting fixes it.

## Why existing recovery is not enough

The current stale-session recovery starts after a known tool call fails. It cannot discover a newly added tool because that tool is not present in Pi's current tool surface, and recovery may happen too late for the first model turn after the server replacement.

## Needed behavior

- Detect a silently stale keep-alive connection or lost notification stream.
- Reconnect with bounded backoff.
- Run a fresh `tools/list` and reinstall list-change notifications.
- Refresh direct tools before the next model turn.

The long-lived Pi session should converge without restarting Pi.

dmorn:
## Problem

A long-lived remote MCP server can be replaced or restarted while its keep-alive connection still appears connected. If tool-list notifications are unavailable or lost, the adapter keeps the old session and stale tool catalog, so Pi can expose tools that no longer exist and miss their replacements.

## Changes

- Force an authoritative, cache-bypassing catalog refresh before user input, before adapter-triggered model turns, and during keep-alive health checks.
- Reconcile changed tools in place, and reconnect only after the server proves the current session is expired or closed.
- Publish replacement metadata before the turn continues or a recovered tool call is replayed.
- Bound polling to 5 seconds per request, 10 concurrent servers, and exponential retry backoff; preserve auth and connection identity/race handling.
- Treat polling as the correctness fallback when the notification stream is unavailable or silently lost—the adapter cannot directly observe a dead GET stream.

## Verification

- `npm run typecheck`
- Focused lifecycle, manager, input-barrier, and recovery tests: 138 passed
- Streamable HTTP wire integration tests: 9 passed
- `npm pack --dry-run`
- Full local suite after building the visualizer: 1150 passed; 2 existing macOS helper-cleanup timing tests failed in the untouched `request-headers-command.test.ts` (GitHub Actions runs on Linux)

Fixes #369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Recovered MCP gateway requests nested inside proxy `args` instead of silently showing status, and now rejects invalid nested gateway requests with guidance. Thanks to [@ibrmora](https://github.com/ibrmora) for #363.
+- Refreshed remote keep-alive tool catalogs before user input, adapter-triggered turns, and during health checks, reconnecting expired Streamable HTTP sessions so long-lived Pi sessions can discover replacement catalogs without restarting. Fixes #369.
 
 ## [2.26.0] - 2026-08-14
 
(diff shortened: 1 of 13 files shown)


nicobailon:
Thanks for the keep-alive convergence work. I landed it through #374 with the same implementation rebased onto the current queue plus the required changelog credit for your contribution.

Closing this draft PR as superseded by #374.

PR #377, opened 2026-08-19
fix: normalize tool-call arguments before transport (embedded-quote serialization)

sebbean:
## Problem

Tool-call args containing literal double-quotes in free-text fields (e.g. a mailbox message body with a quoted phrase) could corrupt agent→MCP serialization:

- **proxy-modes.ts / direct-tools.ts** passed `arguments` straight to the SDK `callTool`. An object is JSON.stringify'd correctly, but a **string-typed `arguments` slipped through un-normalized**, breaking serialization and dropping sibling fields (server-side `Required` errors).
- **ui-server.ts** **silently collapsed** string-typed `arguments` to `{}` — losing every field including required ones.

## Fix

Add `normalizeToolArguments` to `utils.ts` and apply it at every `callTool` transport boundary:

- Parses JSON **strings once** into an object (never re-stringifies/concatenates).
- JSON round-trips object args so embedded quotes are always escaped by `JSON.stringify` at transport, never string-interpolated.
- Rejects arrays/primitives with descriptive errors, so the server always receives a plain object map.

## Behavior

- `normalizeToolArguments(undefined | '' | null)` → `{}`
- `normalizeToolArguments('{"a":1}')'` → `{ a: 1 }`
- `normalizeToolArguments({ body: 'He said "hi"' })` → round-trips intact (quotes preserved as data)
- `normalizeToolArguments([1,2])` → throws descriptive error
- malformed JSON string → throws descriptive error

No API change; server contract unchanged.

### DIFF
--- a/direct-tools.ts
+++ b/direct-tools.ts
@@ -15,7 +15,7 @@ import { createToolSelectorCandidateIndex, formatToolName, getToolNameCandidates
 import { isUiToolVisibleToModel } from "./ui-tool-visibility.ts";
 import { resourceNameToToolName } from "./resource-tools.ts";
 import { authenticate, supportsOAuth } from "./mcp-auth-flow.ts";
-import { formatAuthRequiredMessage, resolveServerUrl, truncateAtWord } from "./utils.ts";
+import { formatAuthRequiredMessage, normalizeToolArguments, resolveServerUrl, truncateAtWord } from "./utils.ts";
 import { SessionRecoveryAuthRequiredError, withSessionRecovery } from "./session-recovery.ts";
 import { combineAbortSignals, isAbortError } from "./runtime-owner.ts";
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks for the fix. I opened #378 as a maintainer replacement so we can keep your transport normalization and also address the approval-key mismatch that Greptile found on the same head.

The replacement preserves the behavior from this PR and adds visible changelog credit: `Thanks to [@sebbean](https://github.com/sebbean) for PR #377.`

I will close this PR after the replacement lands.

nicobailon:
#378 has landed the replacement.

It preserves the transport-normalization behavior from this PR, fixes the approval preview/session-key mismatch from the Greptile review, and keeps changelog credit for your contribution.

Thanks again for the fix.

PR #403, opened 2026-08-20
fix(oauth): release callback listener when idle so peer processes can bind

trevorleibert-mixpanel:
## Problem

The OAuth callback listener is bound lazily on the first `/mcp-auth` in a process, but only released via `stopCallbackServer()` at shutdown when `activeRuntimes.size === 0`. So the first Pi to ever do an OAuth flow in a session holds the (potentially pre-registered) callback port for its entire lifetime, blocking every other Pi's `/mcp-auth` with `EADDRINUSE` even though no auth flow is actually pending.

Concrete symptom on a devbox that runs multiple Pi sessions (tmux panes, sub-agents, etc.):

```
$ lsof -iTCP:3118 -sTCP:LISTEN
COMMAND    PID           USER   FD   TYPE   DEVICE SIZE/OFF NODE NAME
pi      899022 trevor_leibert   24u  IPv4 37085205      0t0  TCP localhost:3118 (LISTEN)

$ /mcp-auth slack   # in a different Pi
Error: Failed to authenticate "slack": OAuth callback port 3118 is already in use.
Pre-registered OAuth clients require an exact redirect URI; set
MCP_OAUTH_CALLBACK_PORT to your registered port or free port 3118
```

The port has been held for 1d21h in that example — nothing is pending; the holder just never released after its one auth on Monday.

## Fix

Release the port immediately after every completed / cancelled / removed auth when no other pending or reserved states remain. The next auth flow in this process rebinds on demand via `ensureCallbackServer`, so releasing eagerly is safe and matches the CHANGELOG intent (\"Avoided OAuth callback port exhaustion by starting the callback server lazily\").

Two-layer change:

1. **Two new accessors on `mcp-callback-server.ts`** alongside the existing `isCallbackServerRunning()` / `getPendingAuthCount()`:
   - `getReservedAuthStateCount()` — mirrors `getPendingAuthCount` for reserved states
   - `isCallbackServerIdle()` — true iff both pending and reserved counts are 0
2. **`mcp-auth-flow.ts` uses a wrapper at the true completion sites** (`startAuth` error path, manual-auth timeout cleanup, `completeAuth` finally, cancellation, `removeAuth`):
   ```ts
   async function clearPendingAuthAndReleaseIfIdle(runtime, serverName, oauthState, opts) {
     await clearPendingAuth(runtime, serverName, oauthState, opts)
     if (isCallbackServerRunning() && isCallbackServerIdle()) {
       await stopCallbackServer()
     }
   }
   ```
   Kept `clearPendingAuth` alone at the two paths that should NOT release: `setPendingAuth`'s sweep-before-set (about to add a fresh record) and `shutdownOAuth`'s iteration (`activeRuntimes` check handles teardown itself right after).

## What this does NOT change

- **Behavior when auths ARE pending**: the port stays bound (blocked by `isCallbackServerIdle()` returning false).
- **`stopCallbackServer` semantics**: still idempotent via the existing `stoppingPromise` cache.
- **`MCP_OAUTH_CALLBACK_PORT` env var**: still honored for pre-registered clients.
- **Random-port fallback for DCR clients**: still available via `strictPort: false`.
- **Cross-process contention**: still exists in the (unlikely) case that two Pis do interactive `/mcp-auth` in the same ~30s window. But shrinks the contention window from days to seconds — good enough for a single-human devbox, and any real IPC-based multi-process serialization would be much more invasive.

## Bonus: drop the redundant \"select Yes to paste\" confirm

Second commit drops the pre-existing Yes/No confirm from the remote/headless paste flow in `commands.ts`. The `onAuthorizationUrl` notify already surfaces the URL as a clickable terminal hyperlink, so by the time the input opens the user has already clicked/copied the link. The confirm reads awkwardly (\"select Yes to paste the callback URL\" — the actual verb is paste, not yes) and adds a step. Dropped, paste input opens directly; Enter after pasting completes the flow. Placeholder text slightly clearer.

Happy to split this into a separate PR if you'd rather review them independently — they're semantically distinct even though both live in the same adapter surface. Together the diff is small enough that landing them together felt fine.

## Testing

- `npm run typecheck`: clean
- `npm test`: **1216 passed** (was 1215 pre-change — +1 from the abort-before-input regression, +2 from the release-on-idle tests, -1 assertion for the dropped confirm updated in place)
- New unit tests:
  - `mcp-callback-server-unref.test.ts`: `isCallbackServerIdle` reflects both pending and reserved state
  - `mcp-auth-flow-client-credentials.test.ts`: auth flow releases callback server after completion when no other auths pending; does NOT release when a second auth is still pending
  - `commands-auth.test.ts`: paste input opens directly (no confirm); aborted-before-input path opens neither
- Locally patched at `~/.pi/agent/npm/node_modules/pi-mcp-adapter/` on a Mixpanel devbox with multiple concurrent Pi sessions; awaiting an in-vivo `/mcp-auth` cycle to prove end-to-end port handoff between processes (test suite already exercises the code path directly).

## Refs

Follows PR #330 (v2.23.0, remote/headless OAuth callback pasting) and PR #331 (auth-link visibility). Both merged; this PR closes the last known ergonomic gap in the paste flow.

### DIFF
--- a/__tests__/commands-auth.test.ts
+++ b/__tests__/commands-auth.test.ts
@@ -132,7 +132,7 @@ describe("authenticateServer", () => {
     );
   });
 
-  it("surfaces the OAuth URL as one terminal hyperlink and accepts a pasted remote callback", async () => {
+  it("surfaces the OAuth URL as a terminal hyperlink in the notify, then opens the paste input directly (no redundant confirm)", async () => {
     const authorizationUrl = "https://auth.example.com/authorize?resource=https%3A%2F%2Fmcp.sentry.dev%2Fmcp";
     const callbackUrl = "http://localhost:3118/callback?code=code&state=state";
     const inputController = new AbortController();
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks for this. I landed the accepted callback-listener fix through maintainer PR #406, with your contribution credited in the changelog and commit co-author metadata.

I used a replacement PR because review found a few lifecycle races that needed maintainer rework on top of this branch. The landed version keeps the same user-facing goal: idle OAuth flows release the callback listener so peer Pi processes can bind it.

Closing this as superseded by #406.

PR #407, opened 2026-08-20
fix: suppress Node 24 mcpScript FD warnings

blalor:
## Why

On Node 24, repeatedly running `mcpScript` emits warnings such as:

```
Warning: File descriptor 22 closed but not opened in unmanaged mode
Warning: File descriptor 22 opened in unmanaged mode twice
```

Each script runs in a short-lived worker. Node's unmanaged file descriptor bookkeeping races with worker termination, even though the worker sandbox cannot open files through `process`, `require`, or other host globals.

## What changed

- Disable unmanaged file descriptor tracking for the `mcpScript` worker. The host still explicitly terminates every worker.
- Add a Node 24 regression test that runs 32 workers in a subprocess and rejects unmanaged file descriptor warnings on stderr.
- Run the focused worker test under Node 24 in CI. The existing CI job remains on Node 22.

Without the worker option, the new regression test consistently fails on Node 24.16.0. With it, the focused test and the full 1,217-test suite pass without the warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,3 +31,17 @@ jobs:
       - run: npm run test:oauth
       - run: npm run test:conformance
       - run: npm pack --dry-run
+
+  node24-worker-fds:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          cache: npm
+
+      - run: npm ci
+      - run: npm test -- --run __tests__/mcp-code.test.ts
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks for this. I landed the accepted Node 24 `mcpScript` worker warning fix through maintainer PR #409, with your contribution credited in the changelog and commit co-author metadata.

I used a replacement PR because review found one focused test gap: the regression needed to reject both unmanaged-FD warning forms. The landed version keeps the worker change and adds that matcher fix.

Closing this as superseded by #409.

blalor:
Thanks @nicobailon ! I’m sure you can tell I put no actual effort into this, just let the clanker rip and hoped it’d be useful. Thank you for your work!

PR #404, opened 2026-08-20
fix(oauth): hide MCP picker during nested auth UI

trevorleibert-mixpanel:
## Summary

- hide the `/mcp` or `/mcp-auth` picker while OAuth shows its authorization link and callback paste input
- restore and refocus the picker after success or failure
- cover both picker entrypoints and failure restoration

## Stack

Depends on #403. This branch includes #403 through a merge commit so the callback-listener lifecycle fix remains independently reviewable and mergeable first.

## Validation

- `npm run typecheck`
- focused command and callback-listener tests (29 passed)
- pre-stack UI branch: `npm test` (1,214 tests)
- pre-stack UI branch: `npm run test:oauth` (137 tests)

PR #408, opened 2026-08-20
fix(oauth): recover fresh Linux keyrings

trevorleibert-mixpanel:
## Summary

- recover missing, expired, rejected, and revoked Linux session keyrings through the existing `keyctl` helper
- preserve the original secure credential-store error when OAuth startup fails instead of masking it with `OAuth startup cleanup failed`
- treat Bearer-challenged HTTP 401 responses as authentication-required MCP endpoints even when the unauthenticated body is not JSON-RPC

## Context

Fresh/headless Linux sessions can inherit an unavailable session keyring, especially through long-lived shells or tmux. Several valid hosted MCP endpoints also return plain JSON or text 401 bodies before authentication. Together these failures made first-time authentication fail with either an aggregate cleanup error or the misleading `this URL does not appear to speak MCP` diagnostic.

Credential handling remains fail-closed. Recovery is limited to concrete session-keyring lifecycle failures and still requires `keyctl` plus `node` on `PATH`; the adapter never falls back to plaintext credentials.

This is stacked on #404 for the remote authentication UI changes. #403 is already on `main`.

## Validation

- `npm run typecheck`
- `npm run test:oauth` — 137 passed
- `npm test` — 1,223 passed after building the interactive visualizer fixture
- focused Vitest suite — 69 passed
- live unauthenticated probes against Linear, Notion, Chronosphere, and Mixpanel classify each Bearer-challenged 401 as authentication-required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reused MCP panel direct-tool counts and token totals across renders while keeping toggle and reconnect updates immediate.
 
 ### Fixed
+- Recovered missing, expired, and rejected Linux session keyrings through the existing `keyctl` helper, preserved actionable secure-store startup errors during OAuth cleanup, and stopped labeling Bearer-challenged 401 endpoints as non-MCP.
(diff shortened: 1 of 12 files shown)


trevorleibert-mixpanel:
Closing at the request of the pilot owner; this recovery change is not needed.

PR #422, opened 2026-08-21
fix: preserve OAuth credentials across processes

mjlbach:
## Summary

- keep SDK token and dynamic-client invalidation local to the provider that observed the rejected credentials
- evict the process-local auth cache after invalidation so stale processes can adopt credentials replaced in the shared OS credential store
- distinguish OAuth refresh reads (`tokens(ctx)`) from transport bearer reads, preserving a newer token saved while an older refresh was in flight
- clear retained credential identifiers when the provider is deactivated

## Problem

OAuth credentials are shared through the OS credential store, while each Pi process has its own auth cache. A long-lived process can therefore retain an old refresh token after another process completes reauthorization.

When the old refresh receives `invalid_grant`, the MCP SDK invokes `invalidateCredentials()`. The provider previously deleted tokens (and, for `invalid_client`, client information) using a cached read/modify/write of the shared credential entry. That stale process could erase the newly authorized token or overwrite the entire newer entry, causing OAuth to appear to expire immediately after the user fixed it.

This was observed with two Pi sessions using the Pylon MCP server: the shared Keychain entry was rewritten to retain only client registration at the exact time the stale session attempted its next call.

## Approach

SDK invalidation now rejects the observed token/client only inside that provider instance rather than mutating shared storage. While a value is locally invalidated, reads bypass the process cache:

- if secure storage still contains the rejected value, the current flow treats it as absent and can reauthorize;
- if another process has replaced it, the stale provider adopts the replacement without deleting it.

Explicit `all` invalidation/logout still removes shared credentials as before.

## Tests

- `npm run typecheck`
- `npm run test:oauth` — 140 passed
- `npm test` — OAuth changes pass; full suite reached 101/102 files, with the pre-existing timing-sensitive `request-headers-command` helper-marker test intermittently failing under full parallel load. The test file passes alone: `npx vitest run __tests__/request-headers-command.test.ts` (18/18).

Added focused regressions for:

- provider-local token invalidation
- adoption of a token replaced by another process
- delayed invalidation of an older token after a newer token is saved
- provider-local dynamic-client invalidation
- adoption of replacement client information without overwriting replacement tokens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- OAuth token invalidation now preserves credentials replaced by another Pi process instead of letting a stale refresh delete newly authorized shared credentials.
 - HTTP 202 and unauthenticated HTTP 401 endpoint probes now report ambiguous endpoint shape instead of claiming the URL is not MCP. Thanks to [@jayshah5696](https://github.com/jayshah5696) for #415.
 - Expanded `mcpScript` calls now show bounded submitted code. Thanks to [@DenisBalan](https://github.com/DenisBalan) for #413.
 - Gateway parameters nested inside `args` now fail with top-level guidance instead of dispatching inconsistently. Thanks to [@voidfreud](https://github.com/voidfreud) for PR #417.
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks again for the fix, @mjlbach.

I landed this through maintainer replacement PR #423 so I could add the follow-up token-identity fix from fresh review and keep the changelog credit in the same branch. The preserved behavior is the same core contribution from this PR: stale Pi processes no longer delete OAuth credentials replaced by another process.

Closing this PR as superseded by #423.

PR #424, opened 2026-08-21
Clarify transient MCP startup failures

elkaix:
## Summary

- classify direct and one-level wrapped HTTP 503 connection failures through one shared predicate
- report startup 503s as a quiet, actionable temporary-availability warning instead of a hard endpoint/protocol failure
- preserve the existing hard failure path for non-transient errors
- preserve documented lifecycle semantics: only `keep-alive` converges automatically; `eager`, `lazy`, and pre-first-connect `lazy-keep-alive` behavior is unchanged
- keep one Pi connection attempt per gateway attempt; the gateway remains responsible for immediate ride-through retries

This is a focused follow-up to merged PR #411. There were no open fork PRs to supersede or close; this branch is newly based on current upstream `main` and contains one clean commit.

## Validation

- `npm run typecheck` — passed
- `npx vitest run --maxWorkers=1` — 1266 passed
- focused transient/lifecycle suites — 67 passed
- `git diff --check origin/main...HEAD` — passed

## Regression coverage

- direct HTTP 503
- wrapped HTTP 503 cause
- non-503 hard failure
- keep-alive startup recovery without changing eager/lazy contracts
- test runtime-owner and temporary-directory cleanup


nicobailon:
Thanks again for PR #424. I landed the maintained replacement in #426.

The replacement keeps your transient startup warning work, preserves your visible changelog credit, and adds the nested-cause fix that came out of review so the real manager-enriched HTTP 503 path also stays on the temporary warning path.

Closing this PR as superseded by #426.

PR #462, opened 2026-08-27
feat: support a single MCP ingress with strict lazy preflight

POLSKA-GRUPA:
## Summary
- add backwards-compatible `namespaceProxyTools` opt-out for one model-facing `mcp` gateway
- add strict cached-schema argument validation before lazy server connection
- preserve canonical `server/tool` resolution across describe and reconnect paths
- add opt-in metadata-only routing telemetry with fail-closed persistence sanitization
- fix macOS request-header helper cleanup discovered by the full suite

## Compatibility
Both new routing settings default to the existing behavior. No server connection occurs for strict preflight failures.

## Validation
- 109 test files
- 1,354 passed, 1 skipped
- typecheck, public build, and interactive visualizer build passed
- provider canary: 100/100 representative routing cases across GLM and OpenAI

The implementation is currently pinned by SHA in a downstream canary and has completed independent cross-review.


nicobailon:
Thanks for the careful work here. I reviewed the exact head and I am going to close this PR rather than keep the broad branch open.

There are useful ideas in it, but it needs to be split before it can land. The current head is also conflicting with `main`, Greptile has a live validation-cost finding, and the routing trace code is not wired into a production call path.

The parts that are worth revisiting as separate PRs are:

- non-blocking request-header cleanup tracking;
- a focused single-ingress or namespace-proxy opt-out change;
- strict cached-schema preflight with bounded validation work;
- routing trace emission, if it has a clear consumer and production emission path.

If any of these pieces land later, I will keep credit for this PR attached to the relevant change. Please keep each follow-up small enough to review and test on its own.

PR #479, opened 2026-09-01
feat(panel): enable/disable MCP servers from the /mcp panel

ericykim:
## Summary

The `/mcp` panel could toggle tools between direct/proxy modes, but enabling or disabling a server required leaving the panel and running `/mcp disable <server>` + `/reload`. This adds the same capability in-panel:

- `ctrl+d` on a server row toggles its disabled state (shown in the hints bar)
- Disabled servers render the existing `disabled` status label
- Saving persists only the `disabled` field via the existing `writeProjectServerDisabledOverride` writer (project Pi layer, never copies credentials)
- If anything changed, the session reloads automatically — no manual `/reload` needed

## Changes

- `types.ts`: `McpPanelResult` gains `disabledChanges: Map<string, boolean>`
- `mcp-panel.ts`: track `disabled`/`wasDisabled` per server, `ctrl+d` toggle, dirty tracking, hint line, disabled-change reporting in the save result
- `commands.ts`: `openMcpPanel` persists disable toggles via `writeProjectServerDisabledOverride` and signals `configChanged` so the session reloads
- Tests: new `mcp-panel-disable.test.ts` (5 tests covering toggle, save, discard, re-enable, disabled rendering); updated the save-result assertion in `mcp-panel-keybindings.test.ts`

## Test plan

- [x] `npm run typecheck` clean
- [x] Panel + disabled-server tests: 60 passed
- [x] Full `npm test`: 1390 passed; only pre-existing failures remain (`interactive-visualizer-server`, `request-headers-command` — both fail on clean `main` too)


ericykim:
Fixed in e41ce67 — `ctrl+d` now requires `item.type === "server"`, so selecting a tool row and pressing `ctrl+d` is a no-op. Added a regression test (`ctrl+d on a tool row does not toggle the server`) that expands a server, moves the cursor to a tool row, toggles, and asserts empty `disabledChanges` on save.

PR #494, opened 2026-09-03
fix(direct-tools): report connect-discovered direct tools via addedToolNames

### LINKED ISSUE #490: Use addedToolNames for result-scoped direct-tool activation
## Summary

Direct-tool activation after `mcp({ connect })` should use Pi's result-scoped tool activation surface instead of relying on active-tool list rewrites.

## Current code

- `index.ts` calls `syncToolSurface()` after `executeConnect()`.
- Direct-tool sync registers new direct tools, calls `pi.setActiveTools(...)` in fallback paths, and carries `fallbackDeactivatedTools` for hosts without `unregisterTool`.
- `/Users/nicobailon/Documents/development/pi-mono/packages/agent/src/types.ts` defines `AgentToolResult.addedToolNames`, and the agent loop persists it onto `ToolResultMessage`.
- Pi's deferred-tools support uses `addedToolNames` as the transcript point where newly discovered tools become available.

## Desired shape

When a connect/search path discovers or refreshes direct tools, return the added names on the tool result via `addedToolNames`. Keep removal/deactivation behavior explicit and fail-closed, but stop using whole-list rewrites just to surface additions.

## Acceptance criteria

- A successful `mcp({ connect: ... })` that discovers direct tools returns those tool names in `addedToolNames`.
- If a search/list path is intended to activate direct tools, it uses the same result-scoped added-tool mechanism.
- Existing active tools are not reordered or rewritten solely to add MCP direct tools.
- Hosts without `unregisterTool` still do not leave stale tools callable after config changes.
- Tests prove new direct tools are available from the result turn and stale/removed tools are handled safely.

## Notes

Estimated deletion: 50-80 lines, depending on how much fallback activation code remains necessary for removals.

chiptoe-svg:
#490 asks that a connect which discovers direct tools report them on the result via `addedToolNames` instead of relying on active-list rewrites, and estimates that 50-80 lines of fallback activation could then go.

This PR does the first part: `mcp({ connect })` now returns newly registered direct tools in `addedToolNames`, and nothing in that path rewrites the active list. It deletes nothing, because on Pi the fallback code is still load-bearing.

The names come from a before/after snapshot of `registeredDirectTools` around `executeConnect`, since a live connect syncs the tool surface from the metadata-update hook before the connect branch's own sync runs. New tools are activated by registration alone: Pi's `_refreshToolRegistry` appends names new to the registry to the active set.

Why the fallback stays: `addedToolNames` is a load-point marker, not an activator. Pi has no `unregisterTool`, so a removed direct tool stays registered, and when it comes back re-registration is not "new" and Pi does not re-activate it. The explicit re-add is what makes it callable again; removing that code would leave returning tools registered but silent. The returning name is still reported in `addedToolNames`, so the signal is consistent both ways.

Typecheck and vitest pass; the two failures in `interactive-visualizer-server.test.ts` are pre-existing on a clean `main`. The two new tests fail without the `index.ts` change.

Closes #490


chiptoe-svg:
Addressed the overlap finding in 1ae959a: each registered direct tool now records its server, and the connect result reports only names belonging to the server being connected, so another server's refresh or a concurrent connect landing mid-await is no longer attributed here. Test added for that case. One caveat outside the adapter: Pi's extension wrapper also diffs the active set around any tool execute and unions that into addedToolNames, so a tool activated by an unrelated refresh during the call can still be attributed by Pi itself.

nicobailon:
Thanks [@chiptoe-svg](https://github.com/chiptoe-svg) for the contribution here. I landed this through the maintainer replacement in #502, which keeps the `addedToolNames` behavior and includes the follow-up concurrency fix needed for overlapping connects. Closing this PR as superseded by #502.

PR #493, opened 2026-09-03
feat: load local Claude plugin bundles

gugu91:
## Summary

- add explicit `claudePlugins` configuration for trusted local Claude plugin directories
- independently load root `.mcp.json` defaults and discover `skills/**/SKILL.md` files at startup and reload
- validate optional `.claude-plugin/plugin.json` manifests, MCP server definitions, paths, symlink containment, and namespace conflicts
- keep package, Agent Plugin, and normal Pi MCP configuration above Claude plugin defaults
- document configuration, precedence, reload behavior, and the local trust boundary

## Scope / trust model

This only reads directories explicitly listed in adapter configuration. It does not download plugins, execute hooks, load commands/agents, or fetch skills through MCP instructions. Manifest component overrides are validated but intentionally not followed; only root `.mcp.json` and root `skills/**/SKILL.md` are supported.

## Validation

- `npm run typecheck`
- `npm run test:public-exports`
- `npx vitest run --exclude __tests__/request-headers-command.test.ts` — 113 files / 1400 tests passed
- focused Claude plugin, lifecycle, runtime registration, config, namespace, and real reload suites passed
- `npm pack --dry-run` includes source and generated Claude plugin loader files
- code-reviewer follow-up: approved, no remaining findings

`__tests__/request-headers-command.test.ts` has pre-existing process-cleanup timing failures on this macOS host: upstream `main` reproduced 4 failures in an isolated worktree, while the feature branch reproduced 5 in one run. This change does not touch that process-management path.

PR #504, opened 2026-09-04
feat(claude-plugins): load trusted local bundles

nicobailon:
## Summary
- loads trusted local Claude plugin bundles from configured directories
- imports bundled MCP server defaults and skills while preserving normal Pi config precedence
- normalizes programmatic relative plugin paths once at the `createMcpAdapter({ config })` boundary so early and runtime setup use the same bundle
- keeps local path trust validation, package artifacts, and resource reload behavior covered

Supersedes #493.

Thanks to [@gugu91](https://github.com/gugu91) for PR #493.

## Validation
- `npx vitest run __tests__/index-lifecycle.test.ts __tests__/claude-plugin-loader.test.ts __tests__/claude-plugin-resource-discovery.test.ts __tests__/runtime-register.test.ts`
- `npm run typecheck`
- `npm run test:public-exports`
- `npm pack --dry-run --json`
- `git diff --check`


will-jack-and-jilll:
Thank you very much @nicobailon really appreciate the responsiveness here! My agents decided to upstream our first draft instead before I even had a chance to review, but it was likely the right move as we got a much better version from you!

PR #519, opened 2026-09-05
fix(namespace): correct tool describe guidance

r1ckyIn:
## Why

Namespace proxy argument descriptions currently tell agents to inspect a tool with `mcp({ describe: 'server/tool' })`. The adapter's describe resolver does not interpret slash-qualified names, so following that instruction produces `tool_not_found` (for example, `linear/get_issue` instead of the discovered `linear_get_issue`). This is adapter-level discovery behavior, not a Linear-specific naming requirement.

## What changed

- Replace the unsupported slash example with search-first guidance: pass the exact name returned by `mcp({ search: ... })` to `mcp({ describe: ... })`.
- Extend the namespace registration test to inspect the actual model-facing argument description and reject the old example.
- Verify search-to-describe discovery for all four `toolPrefix` modes, both globally and with per-server overrides.

## AI involvement

- AI assisted diagnosis, implementation, test generation, and diff review.
- The human requested the upstream fix and approved testing both the registered instructions and the discovery flow.

## Validation

- Confirmed the new instruction regression assertion fails against the original description, then passes after the fix.
- `npm test -- --run __tests__/namespace-tools.test.ts __tests__/proxy-modes-discovery.test.ts` — 59 tests passed.
- `npm run typecheck` — passed.
- `git diff --check` — passed.

## Risks / follow-ups

No resolver, configuration, authentication, or dispatch behavior changes. Tests use local metadata and do not connect to live MCP servers. The full test, OAuth, and conformance suites were not run locally.

### DIFF
--- a/__tests__/namespace-tools.test.ts
+++ b/__tests__/namespace-tools.test.ts
@@ -266,7 +266,7 @@ describe("syncNamespaceProxyTools", () => {
     expect(registered.has("mcp__other")).toBe(false);
   });
 
-  it("exposes a `tool` and optional `args` parameter schema for dispatch", async () => {
+  it("exposes dispatch parameters with search-first describe guidance", async () => {
     const { syncNamespaceProxyTools } = await importSync();
     const { pi, registered } = makePi();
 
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks to @r1ckyIn for the fix and prefix-mode tests. Merged via #523, preserving your contribution with Co-authored-by attribution and a profile-linked Unreleased changelog credit. The final guidance also clarifies discovery availability and unique-name describe behavior. Exact-head CI, review, and post-merge main CI passed. Closing this PR as superseded by the merged replacement; your fork branch is unchanged.

PR #529, opened 2026-09-07
fix: keep namespace proxy tool names under provider length limits

saikpr:
## Problem

`formatServerNamespace` hex-encodes the whole server name when it contains any unsafe character. A dotted name like `awslabs.aws-documentation-mcp-server` becomes a 117-char `mcp__` tool name. Bedrock rejects tool names longer than 64 chars:

```
Validation error: Value 'mcp___mcpns_61_77_73_6c_61_62_73_2e_61_77_73_5f_...' at 'toolConfig.tools.11.member.toolSpec.name' failed to satisfy constraint: Member must have length less than or equal to 64
```

## Fix

Encode only unsafe characters (same scheme as `sanitizeServerPrefix`), keep the `_mcpns_` marker for collision safety.

| Server | Before | After |
|---|---|---|
| `awslabs.aws-documentation-mcp-server` | 117 chars | `mcp___mcpns_awslabs_2e_aws_documentation_mcp_server` (51) |
| `数` | `mcp___mcpns_6570` | `mcp___mcpns__6570_` |

Encoded names for non-ASCII servers change shape; ASCII names without unsafe chars are unchanged. Tests updated, changelog entry added.


saikpr:
Addressed both P1s:

- **Injectivity**: `_` now encodes as `__`, so `__` and `_<hex>_` form a prefix code. `_mcpns_.` → `mcp___mcpns___mcpns___2e_`, `_mcpns__2e_` → `mcp___mcpns___mcpns____2e__`.
- **Length**: namespace capped at 59 chars (64 − `mcp__`). Over-long names get `_mcpns_<head>_<sha256[:16]>`. `数据库数据库数据库` → 64 chars exactly; 70×`a` and 71×`a` stay distinct.

Tests added for both. `dist/types.js` rebuilt via `build:public`.

saikpr:
Digest now hashes the injective ASCII encoding (`namespace`) instead of the raw name, so lone-surrogate vs U+FFFD inputs no longer share bytes. Test added.

saikpr:
@nicobailon  - quick review?

nicobailon (changes requested):
Thanks for addressing the earlier injectivity, length, and surrogate-hashing findings. Reviewing `d8aa1c948c636d922a633d45e606db78b5eac509`, one deterministic namespace collision remains:

- The digest fallback uses the same `_mcpns_` body form as ordinary escaped names. A long safe name can therefore truncate to text that is also the complete encoding of a shorter dotted name, without any SHA collision. For example, with `f = formatServerNamespace`, these distinct names produce the same result:

```js
const a = 'a_2e_' + 'b'.repeat(29) + '_' + 'x'.repeat(30);
const b = 'a.' + 'b'.repeat(29) + '_' + f(a).slice(-16);
f(a) === f(b); // true
```

Both become `mcp___mcpns_a_2e_bbbbbbbbbbbbbbbbbbbbbbbbbbbbb__bb2a80fe8c45d536`. Since that formatter is also the config/reference identity, both server proxies can be suppressed as a collision. Give the hashed form a discriminator that ordinary encoding cannot produce (for example `_h_`, with the retained-head budget adjusted), mirror tracked dist artifacts, and add this pair as a reference-resolution regression.

The focused namespace/reference suites otherwise pass (50 tests), and the patch remains appropriately scoped with contributor credit intact.

saikpr:
Fixed in 20a2d1a: reserved `_h_` discriminator, regression test, and rebuilt dist artifacts. All local checks pass; Greptile: 5/5. @nicobailon, ready for re-review.

nicobailon (approved):
The deterministic namespace collision is resolved with a reserved hash domain. Length, encoding, surrogate, generated-artifact, and reference-resolution coverage pass on the current head; required CI, Node 24 worker-FD coverage, and current-head automated review are green. Contributor authorship and changelog credit are preserved.

PR #532, opened 2026-09-09
feat: install and authenticate MCP servers from one URL

exoulster:
## Summary

- add `mcp({ action: "install", url: "https://example.com/mcp" })` for durable URL-only MCP setup
- validate and provisionally connect endpoints before atomic persistence; roll back invalid endpoints and reject unsafe URLs or name collisions
- reuse existing servers by canonical URL, support explicit names and project/global targets, and avoid a reload for the active proxy surface
- open OAuth authorization through Pi's browser launcher, watch loopback callbacks, complete PKCE token exchange automatically, and emit a turn-triggering completion event
- preserve `auth-complete` for remote/headless and pre-registered HTTPS callback fallbacks

## Intended UX

A user can tell the agent: `Install https://example.com/mcp`. The agent calls the install action. Public servers connect immediately. OAuth servers open consent automatically; after consent the callback watcher returns the agent to connect and verify discovered tools. No manual JSON editing, client ID, callback copy, or second auth command is needed for a reachable local loopback flow.

Registration and authentication remain separate from user authorization: the user still grants consent in the authorization server.

## Verification

- `npm run typecheck`
- `npm run test:oauth` (152 passed)
- `npm test` (1,495 passed)
- `npm run test:public-exports` (2 passed)
- package dry run includes `mcp-install.ts`
- host compatibility checked for `withFileMutationQueue` on Pi 0.84.1 and 0.84.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - HTTPS MCP servers can configure an origin-scoped `caFile` PEM trust bundle without disabling certificate verification, for Streamable HTTP and SSE. Thanks to [@desmonna](https://github.com/desmonna) for #527.
 - `directTools: "search"` registers a server's tools as real direct tools but inactive; `mcp({ search })` activates the matches additively, reported as `addedToolNames`. Activation is per process and search is its only trigger. Search-mode tools do not count toward the 75-tool advisory. Thanks to [@chiptoe-svg](https://github.com/chiptoe-svg) for PR #525.
+- The script MCP gateway can now install, validate, connect, and authenticate a durable server configuration from one endpoint URL with `mcp({ action: "install", url: "..." })`.
 - Script `tools.describe()` now exposes server-advertised output schemas for `data.structuredContent`, preserved through metadata caching and refresh. (#522)
 - Stdio MCP servers can opt out of arbitrary adapter environment inheritance with `inheritEnv: false`; SDK platform defaults and explicit `env` overlays remain. Thanks to [@zenolam](https://github.com/zenolam) for #509.
 - Local Claude plugin bundles can now be loaded from trusted configured directories, including bundled MCP servers and skills. Thanks to [@gugu91](https://github.com/gugu91) for PR #493.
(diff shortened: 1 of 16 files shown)


nicobailon (changes requested):
Thanks for the cohesive URL install/OAuth flow. Reviewing `dbbda54e7f972d914f8dd01586f6010244850eff`, four correctness cases block this head:

1. `target: "project"` succeeds and writes `.mcp.json` even in exclusive config mode, where reload reads only the exclusive config. Reject that destination before provisional runtime work unless it is the active exclusive path.
2. Promoting an existing runtime server validates the full runtime definition but persists only `{ url }`, dropping required headers/OAuth options. Persist the validated definition snapshot or reject runtime promotion rather than claim durable installation.
3. Install discovers/registers direct tools through `executeConnect` but does not return result-scoped `addedToolNames`, unlike the existing connect wrapper. Share that activation/reporting path so successfully installed selected direct tools load at the install result.
4. If persistence fails after successful discovery, rollback removes config/connection but leaves tool/prompt/instruction metadata and provisional cache state discoverable. Remove all provisional metadata/cache state before resynchronizing surfaces.

Please add focused regressions for each before/after behavior. The current focused install/lifecycle/auth suites and typecheck otherwise pass, and the overall feature scope is cohesive; contributor credit remains intact.

nicobailon:
One additional rollback case should be covered with the four requested fixes on `dbbda54e7f972d914f8dd01586f6010244850eff`:

- `executeInstall` inserts the provisional server/lifecycle state before awaiting `executeConnect`. `executeConnect` calls `throwIfAborted` before its own `try`, and `executeInstall` has no exception boundary around that await. An already-aborted tool signal therefore rejects past the rollback and leaves the provisional runtime registration behind even though nothing was persisted.

Please reject cancellation before provisional mutation and guarantee rollback for exceptional connection exits. A focused regression should pass an already-aborted signal and prove that no config/lifecycle registration or write remains. The feature direction and contributor credit remain unchanged.

nicobailon (approved):
The install, persistence, rollback, cancellation, and OAuth ownership findings are resolved on the current head. Focused lifecycle/OAuth validation, required CI, Node 24 worker-FD coverage, and the current-head automated review all pass. Contributor authorship and changelog credit are preserved.

PR #550, opened 2026-09-11
Fix caFile connections on Node 26 via bundled undici fetch

### LINKED ISSUE #549: caFile connections fail on Node 26: bundled undici v6 Agent passed to global (undici v8) fetch
## Summary

Every `caFile` connection fails on Node 26 with `UND_ERR_INVALID_ARG: invalid onError method`, so per-origin custom CA trust (Streamable HTTP, SSE, OAuth metadata) is completely broken on current Node.

## Environment

- pi-mcp-adapter 2.33.0
- Node v26.8.2 (built-in undici 8.10.2) — works on Node 22 (built-in undici 6)
- Dependency `undici 6.28.1`

## Repro

```bash
npm install
npx vitest run __tests__/http-ca.test.ts -t "fails by default"
```

```bash
FAIL  __tests__/http-ca.test.ts > per-origin custom CA > fails by default, ...
Caused by: InvalidArgumentError: invalid onError method
Serialized Error: { code: 'UND_ERR_INVALID_ARG' }
```

9 of 15 tests in `__tests__/http-ca.test.ts` fail the same way.

## Root cause

`createCaFetch` (`http-ca.ts`) builds an `Agent` from the bundled `undici@6` but hands it as `init.dispatcher` to the **global** `fetch`, whose internals are Node 26's bundled undici v8. The two undicis disagree on the dispatcher hook contract (`onError` vs `onResponseError`), so dispatch validation rejects the v6 Agent before TLS even starts. Same failure class as qdrant-js#134 (version matrix there shows only same-major Agent+fetch pairing works, so a dep bump alone just relocates the breakage).

```mermaid
flowchart LR
    subgraph broken ["Broken (Node 26)"]
        A1["Agent<br/>(bundled undici v6)"] -- "dispatcher" --> F1["globalThis.fetch<br/>(built-in undici v8)"]
        F1 -- "UND_ERR_INVALID_ARG<br/>invalid onError method" --> X1(["all caFile requests fail"])
    end
    subgraph fixed ["Fixed"]
        A2["Agent<br/>(bundled undici v6)"] -- "dispatcher" --> F2["undiciFetch<br/>(bundled undici v6)"]
        F2 -- "same-undici pairing" --> OK2(["TLS + hostname/expiry<br/>verification intact"])
    end
```

## Suggested fix

Route same-origin requests through `fetch` imported from the same bundled `undici` copy that provides the `Agent` (per the [undici best practice](https://github.com/nodejs/undici/blob/main/docs/docs/best-practices/undici-vs-builtin-fetch.md): import fetch from undici to use the installed version instead of the built-in). Off-origin requests keep using global fetch. PR follows.

zhulinchng:
Fixes #549.

## Problem

`createCaFetch` built its `Agent` from the pinned `undici@6` dependency but passed it as `init.dispatcher` to the global `fetch`, whose internals track the runtime Node (undici v8 on Node 26). The cross-version dispatcher handoff fails validation (`UND_ERR_INVALID_ARG: invalid onError method`) before TLS starts, so every `caFile` connection — Streamable HTTP, SSE, fallback, and connection-owned OAuth metadata — fails on Node 26. Works on Node 22 only because both sides happen to be undici v6.

## Fix

Same-origin requests now go through `fetch` imported from the same bundled `undici` copy that provides the `Agent`, keeping both sides on one implementation regardless of the Node version (a dep bump alone would just relocate the breakage across Node lines — see the version matrix in #549). Off-origin requests still use the global fetch; redirect rejection, origin scoping, and dispatcher cleanup are unchanged.

## Verification

- Without the fix on Node v26.8.2: 9 of 15 tests in `__tests__/http-ca.test.ts` fail with `UND_ERR_INVALID_ARG`.
- With the fix: 15/15 pass, `npm run typecheck` clean, full vitest suite 1602/1602, `test:oauth` 186/186, `test:public-exports` 2/2, `bash conformance/run.sh` all PASS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Per-origin `caFile` trust now routes same-origin requests through the bundled undici fetch so the custom CA dispatcher matches the fetch implementation on newer Node releases (Node 26 ships undici v8 while the dependency pins undici v6); previously every `caFile` connection failed with `UND_ERR_INVALID_ARG`.
+
 ## [2.33.0] - 2026-09-10
 
 ### Highlights
--- a/http-ca.ts
+++ b/http-ca.ts
@@ -1,6 +1,6 @@
 import { readFileSync } from "node:fs";
 import { X509Certificate } from "node:crypto";
-import { Agent } from "undici";
+import { Agent, fetch as undiciFetch } from "undici";
 import type { ServerEntry } from "./types.ts";
 import { getMissingEnvVars, resolveConfigPath, resolveServerUrl } from "./utils.ts";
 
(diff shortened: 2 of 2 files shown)


nicobailon:
Superseded by the reviewed and merged maintainer promotion #557. Zhu Lin’s original commits and contribution credit were preserved there. Thank you, @zhulinchng.

PR #556, opened 2026-09-12
feat(config): add opt-in ancestor MCP config roots

nicobailon:
Supersedes #555 with a maintainer-owned promotion after `main` advanced. John Henao's original commits and authorship are preserved.

This revision keeps ancestor MCP configuration explicit and bounded:
- adds `settings.ancestorConfigRoots` as an opt-in
- accepts the setting only from user-global or explicitly selected config
- performs no ancestor traversal by default
- selects the deepest configured root containing the current directory
- loads ancestor `.mcp.json` and Pi config files farthest-first, never above that root or `$HOME`
- preserves canonical-alias precedence, provenance, merge behavior, and URL-bound credential stripping

Example:

```json
{
  "settings": {
    "ancestorConfigRoots": ["~/work"]
  }
}
```

Project and ancestor config files cannot enable or widen ancestor discovery.

Validation at `e7977936925040479e12a4c731e23c94b07ca975`:
- 91 focused config/panel tests pass
- typecheck passes
- public build and export tests pass
- regenerated output is deterministic
- fresh review found no P0/P1/P2 issues or required simplifications

Thanks to @johnhenaot for the original implementation in #555.


nicobailon:
The reported startup path is valid. Ancestor discovery extends an existing cwd configuration-trust gap, and tool-call approval does not protect server startup or connection-time helpers.

I’m holding this PR while we add a shared trust policy for cwd and ancestor inputs. Simply checking Pi’s project-trust boolean is insufficient: MCP-only directories can be reported as trusted without a prompt or saved decision. The prerequisite will respect host denials, reuse covering saved grants, require explicit authorization for uncovered configuration sources, and fail closed noninteractively.

Once that prerequisite is reviewed and merged, this PR can retain its intended discovery and precedence behavior with the same protection applied to ancestors.

nicobailon:
The earlier trust hold is resolved by narrowing the feature rather than adding a broad trust subsystem. Ancestor discovery is now disabled by default and can only be enabled with bounded `ancestorConfigRoots` from user-global or explicitly selected configuration. Project and ancestor files cannot enable or widen the search. Exact-head CI and Greptile review pass.

Issue #375, opened 2026-08-18
After upgrading Node (via home-brew) macOS makes me put my Keychain password in a bunch of times


joeldrotleff:
Hello!
I'd like to report this issue - I had to put in my system password several times yesterday after Homebrew updated Node.js (presumably used by pi-mcp-adapter), seemingly once per MCP I had enabled.

Not sure if there is an easy workaround but it would be lovely if I didn't have to do this. Cheers

nicobailon:
Thanks for the report. Current `main` already has process-local OAuth credential caching from #335, so the adapter should do at most one OS credential-store read per distinct OAuth server per adapter process unless the server later returns an OAuth 401.

A Homebrew Node replacement can still make macOS Keychain ask once for the new Node executable. I do not see a current adapter-side path that would repeatedly read the same stored OAuth record after that first access.

I am closing this as already fixed/current behavior. If this still happens on a release that includes #335, please reopen with the adapter version, the exact Keychain prompt application name, and whether the affected MCP servers are OAuth, bearer-token, or stdio servers.

PR #560, opened 2026-09-12
fix(oauth): skip Windows credential chunking on macOS and Linux

jploskonka:
## Problem

OAuth records larger than 1000 characters are stored as a Keychain/credential-store **manifest plus digest-addressed chunks** on every platform. That threshold exists for Windows Credential Manager (`CRED_MAX_CREDENTIAL_BLOB_SIZE` / UTF-16).

On macOS it is harmful:

- Atlassian MCP tokens are ~8–10 KiB, so one server becomes **8–9 Keychain items**.
- Each item prompts separately.
- Each write hashes the payload into the account name (`account.chunk.<digest>.N`), so Always Allow on chunk 3 does not apply to chunk 4 or to the next save.

Related: #375 was closed as “one read per server via cache.” Cache does not help when one record is many items, or when the next write mints new item names.

## Change

- Chunk only on Windows and the `sizelimited` test store.
- macOS Keychain / Linux Secret Service keep a single item unless the payload is pathologically large (32k chars).
- On the next non-status read, rewrite an existing chunked record as one item when it now fits. Status-panel inspection does not compact.

## Tests

`npx vitest run __tests__/mcp-auth-storage.test.ts` — 19 passed on darwin.


nicobailon:
Superseded by #565, now merged on `main`. The replacement preserves your original commit and authorship, incorporates the review fixes on current `main`, and credits you in the changelog. Thanks @jploskonka for the report and implementation.

PR #562, opened 2026-09-13
fix: use live metadata for zero-TTL direct tools

### LINKED ISSUE #561: directTools never register when the server advertises ttlMs: 0
I was trying to get some of the GitHub MCP server's tools to be direct tools, but apparently because it advertises `ttlMs: 0`, the current implementation considers the tool metadata immediately invalid, and thus does not cache the tool list, and therefore does not allow you to select certain tools as direct tools. The rest of this issue is GPT-5.6 Sol's output:

> ## Environment
> 
> - pi-mcp-adapter: 2.33.0
> - Pi: 0.85.1
> - Node.js: 24.21.0
> - Server: GitHub Copilot MCP (`https://api.githubcopilot.com/mcp/`)
> 
> ## Configuration
> 
> ```json
> {
>   "mcpServers": {
>     "github": {
>       "url": "https://api.githubcopilot.com/mcp/",
>       "auth": "bearer",
>       "bearerToken": "!gh auth token",
>       "lifecycle": "eager",
>       "directTools": [
>         "get_file_contents",
>         "search_code"
>       ]
>     }
>   }
> }
> ```
> 
> ## Steps to reproduce
> 
> 1. Configure the GitHub server with one or more `directTools`.
> 2. Start Pi.
> 3. Run `/mcp reconnect github`.
> 4. Confirm that GitHub connects and its tools are listed.
> 5. Inspect Pi's active tools/system prompt.
> 
> ## Actual behavior
> 
> - `/mcp` reports GitHub connected with 48 tools.
> - The tools work through the `mcp` proxy.
> - No `github_*` direct tools are registered.
> - Reconnecting does not report any `addedToolNames`.
> - The cache entry contains:
> 
> ```json
> {
>   "ttlMs": 0,
>   "tools": 44,
>   "resources": 4
> }
> ```
> 
> ## Expected behavior
> 
> `ttlMs: 0` should prevent persistent metadata reuse, but authoritative metadata obtained from the current live connection should still be usable to register or hot-load direct tools for the current Pi session.
> 
> Alternatively, the adapter should report explicitly that direct tools cannot be registered for a server with non-cacheable metadata.
> 
> ## Suspected cause
> 
> Since 2.28.0, `isServerCacheValid()` honors server TTL hints:
> 
> ```ts
> if (declaredTtlMs === 0) return false;
> ```
> 
> Direct-tool synchronization appears to reload the persistent metadata cache:
> 
> ```ts
> const cache = loadMetadataCache();
> const result = syncDirectTools(config, cache);
> ```
> 
> `resolveDirectTools()` then skips the GitHub entry because `isServerCacheValid()` returns false. This also happens after a successful live refresh because the refreshed entry retains `ttlMs: 0`.
> 
> As a result, the live metadata is available to proxy operations but never reaches `registerTool()`.
> 
> ## Suggested direction
> 
> Keep treating `ttlMs: 0` as non-cacheable across sessions, but allow direct-tool synchronization to consume authoritative in-memory metadata from an active connection. This should preserve the documented current-session hot-loading behavior without violating the server's cache hint.
>

dsluo:
Closes #561.

Uses freshly loaded tool metadata from MCP servers when synchronizing direct tools, even when server advertises `ttlMs: 0`. Existing zero-TTL metadata remains non-cachable.


nicobailon:
Superseded by #566, now merged on `main`. The replacement preserves your original commit and authorship, incorporates the lifecycle and cache-validity fixes on current `main`, and credits you in the changelog. Thanks @dsluo for the report and implementation.

PR #567, opened 2026-09-13
fix: treat blank JSON files as absent

RobertoNegro:
## Problem

When running under the bubblewrap sandbox, `.mcp.json` is blacklisted so the LLM cannot write to it. If no project-level `.mcp.json` exists, bubblewrap creates the file (empty and non-overwritable) to prevent the LLM from creating it. The adapter then fails while parsing the project-level JSON config because the file is empty (`Unexpected end of JSON input`).

## Change

`parseJsonWithComments` now returns `undefined` for blank input (empty, whitespace-only, or comments-only) instead of throwing, and every call site treats that as "file absent":

- imported host configs: a blank file falls through to the next candidate
- validated/raw/project-override configs: blank behaves like a missing file
- Claude/agent-plugin and package MCP configs: blank means no servers, blank manifests mean no plugin
- legacy plaintext OAuth credentials: a blank `tokens.json` means no credentials

Non-blank behavior is unchanged, including errors on malformed JSON.

## Tests

- `__tests__/utils-parse-json.test.ts`: blank inputs return `undefined`, commented/trailing-comma JSON still parses, invalid JSON still throws
- `__tests__/agent-plugin-blank-config.test.ts`: blank `mcp.json` loads as empty with no warning, blank `plugin.json` is invalid without throwing
- `npm run typecheck` clean; related suites pass (`config`, `claude-plugin-loader`, `package-manifest`, `onboarding-state`, `disabled-server`, `exclusive-config`, `mcp-auth-storage`, `metadata-cache-*`, `claude-plugin-resource-discovery`)


nicobailon:
Superseded by #568, now merged on `main`. The replacement preserves your original commit and authorship while narrowing blank handling to optional layered MCP configuration, keeping required manifests and credential files strict. The changelog credits you and #567. Thanks @RobertoNegro for the report and implementation.

PR #571, opened 2026-09-13
feat(oauth): support client ID metadata documents

dsluo:
This PR adds support for CIMD-based authentication for MCP servers, which replaced Dynamic Client Registration (DCR) for auth in the latest MCP 2026-07-28 spec.

Opening this as a draft PR, as there's still a question of how to serve the client metadata URL that is used as the client ID in the CIMD spec. Needs to be a publicly accessible HTTPS URL that contains the json of said metadata:

```json
 {
   "client_id": "https://<url to this json document>",
   "client_name": "Pi MCP Adapter",
   "client_uri": "https://github.com/nicobailon/pi-mcp-adapter",
   "redirect_uris": [
     "http://localhost/callback",
     "http://127.0.0.1/callback"
   ],
   "token_endpoint_auth_method": "none",
   "grant_types": [               
     "authorization_code",
     "refresh_token"
   ],
   "response_types": [
     "code"
   ]
 }
```

For the URL, I'd suggest hosting it on GitHub pages. Using the URL of the raw `raw.githubusercontent.com` apparently does not work, as it serves it as `text/plain` instead of `application/json`, which some IDPs don't like. This PR does not include the scaffolding to actually serve the metadata doc, as that would require some changes in the repo settings.


nicobailon:
Thanks @dsluo — the opt-in `oauth.clientMetadataUrl` direction makes sense, but this should remain draft for now.

For this PR, please keep the scope to an operator-supplied advanced option. Do not add or depend on a project-hosted default metadata URL yet; deploying and owning that HTTPS document, its canonical loopback redirect, MIME/cache behavior, and release ordering is a separate infrastructure decision.

The merge blocker is migration from stored DCR credentials: when CIMD is newly configured, an existing DCR client plus refresh token must get its normal refresh attempt before switching client IDs. Replacing the stored client first can pair the old refresh token with the CIMD URL and force unnecessary reauthorization. Please add coverage for preserving the DCR pair, then transitioning after invalidation.

Please also define and test `clientMetadataUrl + clientSecret` without an explicit `clientId` (prefer rejecting the ambiguous combination, especially for `client_credentials`), correct the docs so URL-only/default Pi still uses DCR unless this option is configured, and add the usual changelog credit. Existing explicit `clientId`, DCR fallback, ordinary OAuth, and logout/generation fences should remain unchanged.

Issue #570, opened 2026-09-13
Agent Plugins: translated values are re-expanded by the native connection runtime


cheetahbyte:
Disclosure and context: I am building a Pi extension for Agent Plugins that delegates MCP connections to this adapter's runtime registration API, so a literal-value guarantee on that API is what I need most. My agent found these issues while I was building it. I checked the cited lines and ran the tests myself.

**Version:** pi-mcp-adapter 2.33.0 (commit `32b67f916745b31a30a80371a86f00f70ba0a561`)
**Introduced by:** #304 (`e0c4482`, "feat: load MCP servers from Agent Plugins"), hardened in #305 (`f6e5a14`). #383 (runtime registration API) routes through the same connection code, so it does not avoid this.
**Spec:** [Agent Plugins 1.0.0](https://github.com/agentplugins/agent-plugins-spec/blob/main/spec/1.0.0.md), sections 4.1, 5.2, 5.4, 7.2.1, 9.1, 9.2.

Findings 1 to 4 and 6 are reproduced with vitest against commit `32b67f9` (macOS, Node 20). Finding 5 is a question about how to read the spec. The tests are at the bottom. Each one fails on current code and asserts what the spec requires.

## 1. Native `${VAR}` / `$env:` / `{env:}` interpolation is applied to plugin values

Spec 7.2.1 and 9.2: expand only `${PLUGIN_ROOT}` and `${PLUGIN_DATA}` in stdio `args`, `env` and `cwd`, exactly once. Everything else stays literal, and HTTP `url` and `headers` are never expanded.

`translateStdioServer()` ([agent-plugin-loader.ts#L191-L227](https://github.com/nicobailon/pi-mcp-adapter/blob/32b67f916745b31a30a80371a86f00f70ba0a561/agent-plugin-loader.ts#L191-L227)) expands plugin placeholders and sets `literalEnv: true`. That flag only gates `resolveEnv()`. The stdio path in `server-manager.ts` then does:

```ts
let args = (definition.args ?? []).map((argument) => interpolateEnvVars(argument));   // L834
const cwd = resolveConfigPath(definition.cwd) ?? this.defaultCwd;                     // L835
```

[...]

PR #572, opened 2026-09-13
fix: preserve built-in Agent Plugin boundaries

nicobailon:
## Summary

- preserve literal Agent Plugin arguments, environment values, working directories, and HTTP/OAuth headers through built-in loading and config overrides
- prevent plugin header values from entering native environment interpolation or `!command` secret execution
- enforce realpath containment for plugin manifests, MCP configuration, relative commands, and working directories
- validate known manifest field types and expand `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` exactly once

This addresses the adapter's existing built-in Agent Plugin support only. It does not add a public literal-values mode or support external plugin loaders. Same-name `PLUGIN_DATA` behavior is unchanged.

Thanks @cheetahbyte for the detailed report and reproductions in #570.

## Validation

- focused changed-seam and provenance tests passed
- OAuth suite: 173 passed
- full suite: 1,670 passed, 1 skipped
- typecheck, package-manifest checks, and public exports passed
- package self-resolution rejects the internal provenance subpath
- generated public declarations expose no provenance mutation API


cheetahbyte:
Thanks, this covers everything in #570 for the built-in loader.
On "does not add a public literal-values mode": that is the part my extension depends on. It translates Agent Plugins itself and registers the servers through the runtime API, and since the marker is Symbol-keyed and the registration path structuredClones the definition, there is no way to opt in from outside. Is a public opt-in on the registration API something you would consider?

nicobailon:
Thanks for confirming the built-in coverage. I am not planning to add a public literal-values mode to the runtime registration API. That API continues to accept pi-mcp-adapter native server definitions and their native interpolation/command-secret semantics; supporting definitions translated by external Agent Plugin loaders would create a separate compatibility contract beyond this adapter’s built-in Agent Plugin integration. This PR intentionally keeps that provenance private.

PR #583, opened 2026-09-15
feat: hide mcp-scripting skill when scriptMode not enabled

zhangyoufu:


nicobailon:
Thanks @zhangyoufu — your contribution has landed through #584, preserving your authored commit and adding the reload-consistency fix, documentation update, and changelog credit needed for the final behavior.

PR #592, opened 2026-09-16
feat: allow disabling namespace proxy tools

k03mad:
## Summary

Add `settings.namespaceProxyTools` (default: true) to let users opt out of per-server `mcp__<server>` wrappers and reduce tool-schema overhead when using the shared MCP gateway.

- Skip namespace-proxy registration when explicitly set to false.
- Reuse existing cleanup to deactivate previously registered wrappers.
- Keep MCP reference resolution consistent with the setting.
- Leave `mcp`, `mcpScript`, and direct tools unchanged.
- Document the setting and regenerate tracked public type artifacts.

Focused follow-up to the namespace-proxy opt-out discussed in #462.
No preflight validation, routing, tracing, or process-lifecycle changes.

## Validation

- 361 targeted tests passed across namespace/reference, config, lifecycle, direct-tool, and proxy discovery/instructions suites.
- 2 public-export tests passed.
- Typecheck, public build, and git diff --check passed.
- Full CI suite not run locally.


nicobailon:
Thanks @k03mad. The namespace proxy opt-out landed in #599, preserving your authored implementation commit and adding visible changelog credit. I’m closing this original PR as superseded by the landed replacement.

PR #608, opened 2026-09-20
fix: refresh bearerToken !command shellout per HTTP request

kesor:
## Problem

For lazy-keep-alive HTTP servers, the StreamableHTTPClientTransport reuses the resolved bearer token across every request. With `bearerToken: "!command"`, that token is resolved once at connect time and frozen in the transport's `requestInit` until reconnect.

When the token expires (e.g. a Cloudflare Access JWT with a 24h session) the next request returns 401. `shouldReconnectAfterRefresh` ignores 401 (it only triggers on terminated sessions or `NotConnected`/`ConnectionClosed`), so the lifecycle manager logs the failure, backs off, and never reconnects. Quitting pi is the only way to recover.

This affects every `auth: "bearer"` setup where the token can rotate while the connection is alive: `bearerToken: "!cloudflared access login --app=..."`, `bearerTokenEnv` pointing at a refreshed secret, or `bearerTokenStore: true` with rotated credentials.

## Fix

1. **Route `!command` bearer tokens through a per-request `FetchLike`** backed by a TTL-cached resolver (`BearerCommandResolver`). The shellout runs at most once per window per server (default 5 minutes, override via `PI_MCP_ADAPTER_BEARER_COMMAND_TTL_MS`). A transient shell failure falls back to the last good token instead of auth-failing every subsequent request.

2. **Treat 401 as a reconnect trigger** in `shouldReconnectAfterRefresh`. Static, env-var, and stored bearer tokens still bake into `requestInit` at connect time; without this hook they stay permanently stuck on the first rotation until pi restart.

## Changes

- `bearer-command-resolver.ts` (new): TTL-cached, in-flight-coalescing resolver
- `server-manager.ts`: eager resolve + per-request `Authorization` injection; `!command` no longer baked into `requestInit.headers`
- `lifecycle.ts`: 401 reconnects, not silent backoff
- `__tests__/bearer-command-resolver.test.ts` (new): 9 unit tests
- `__tests__/server-manager-http-auth.test.ts`: 2 new tests (no-bake + per-request resolution)
- `__tests__/lifecycle-lazy-keep-alive.test.ts`: 1 new test (401 → reconnect)
- `CHANGELOG.md`

## Test plan

- [x] `npm run typecheck` clean
- [x] 71/71 pass in the relevant test files (bearer auth, lifecycle, opencode interpolation, plus new resolver tests)
- [x] Pre-existing failures in `config.test.ts` and `request-headers-command.test.ts` are unrelated (Windows-only `taskkill` paths, NixOS `recheck-linux-x64` missing); confirmed by `git stash` of my changes
- [ ] Manual smoke test against a real Cloudflare-Access-protected MCP server after merge

## Backwards compatibility

- Default TTL is 5 minutes. With `cloudflared access login`, hits are served from the local JWT cache in <100 ms and re-emit the cached JWT; the user-visible cost is one extra shell process every 5 minutes.
- For servers that set a static `bearerToken: "literal"` or `bearerTokenEnv`, behavior is unchanged: the token still bakes into `requestInit` once, but now the lifecycle manager will reconnect when it expires instead of staying in failure backoff.
- The 401 reconnect only fires for keep-alive refresh errors, not initial connect errors (those still go through the OAuth / `needs-auth` path).


nicobailon:
Superseded by #615, which preserves your original implementation as its first authored commit and adds the maintainer hardening needed for safe landing. Thank you, @kesor, for the design and contribution.

PR #619, opened 2026-09-20
[pi-mcp-adapter] Preserve abort signals after garbage collection

zaini:
HTTP requests using `requestHeadersCommand` can stop responding to cancellation after the wrapper's temporary `Request` is garbage-collected, leaving response streams and sockets open.

Preserve the caller's signal for header generation and fetch, including `Request` inputs, while keeping explicit `signal: null` behavior. Add streaming/GC regressions and signal-override coverage.

Both streaming regressions fail before the fix and pass after it. Typecheck passes; nine focused tests pass on Node 22, 24, and 26. Local validation stubs `ps` because the sandbox blocks process enumeration; HTTP, garbage collection, and cancellation are real. The committed tests use normal process cleanup.


nicobailon:
Thanks @zaini — your abort-signal fix and regression coverage landed through maintainer replacement #625, including visible changelog credit. Closing this original PR as superseded by the landed replacement.

